PR #49, opened 2026-07-11
Add OpenRung Volunteer desktop app (one-click home relay, all 3 OSes)

openrung-admin:
## What & why

The end goal for volunteering is for **non-technical people to run a relay on their home computer** without Docker or a command line. This adds **OpenRung Volunteer** — a one-click cross-platform desktop app (Wails v2 + React, macOS/Windows/Linux from one codebase, same stack as the desktop client) that does exactly that.

It's a **separate app** (`desktop-volunteer/`), not a mode inside the client, because a relay is a long-running background service with the opposite lifecycle (quitting warns you; closing the window shouldn't drop the people using it) and a different UI.

## How it's built

**Engine extraction (root module, reused by `cmd/volunteer` too):**
- `internal/volunteer/engine` — an embeddable, supervised relay: `Start`/`Stop`, status + identity callbacks, capped-backoff session restarts, registration retry honouring `Retry-After`, and a periodic public-IPv6 recheck that re-registers when a home prefix rotates.
- Pulled the broker HTTP client → `internal/volunteer/broker.go` and the reachability probe → `internal/volunteer/probe.go` out of `cmd/volunteer` package main so both the CLI and the app share them (CLI behaviour unchanged; its tests still pass).
- Added opt-in structured connect/disconnect events to the connection observer and live traffic counters + an `OnDisconnected` hook to the tunnel client. Also fixed a latent bug: tunnel reconnect backoff never reset after a good connection, so flaky wifi converged to 30 s dead time.

**App layer (`desktop-volunteer/`):**
- `volunteerservice/` — the Wails bridge, mirroring the client's `vpnservice` pattern (one bound struct, `Emitter` assigned at startup, a single coalesced state event).
- **Informed-consent gate** enforced in `Start()` (server-side, not just UI): running a relay makes your IP a visible traffic exit, explained Tor-style before the first start.
- Live status: phase, transport (direct/tunnel), relay name, public endpoint, people connected, connections served, data shared, uptime.
- Settings (relay name + regenerate, capacity hints, advanced port/broker/hub) persisted to a config dir **distinct from the client's**, so a volunteer install isn't linkable to a client install. Relay identity (Reality keys) persisted for a stable identity across restarts.
- Quit-while-running confirmation so closing the window doesn't silently drop users.
- xray bundled per-OS exactly like the client bundles sing-box — packaging scripts + a release workflow (`volunteer-v*` tags) that **SHA-256-verifies every download**. `THIRD_PARTY_NOTICES.md` updated for the new xray distribution channel.

## Platform question, answered

**All three, one codebase — via Wails, same as the existing desktop client.** Wails can't cross-compile (it links each OS's native webview), so CI builds on native runners (macos-14 / ubuntu-22.04 / windows-latest). No "which first" tradeoff.

## Verification

- Frontend build + 21 vitest tests pass; `go build` + `go test` (incl. `-race` on engine/tunnel) pass.
- Full UI flow driven in the built app: consent gate → online with live stats → settings-lock-while-running → about.
- The embedded engine **registered through a real relay hub into a real broker with bytes round-tripping end-to-end** (tunnel mode, echo backend).
- A multi-agent adversarial review found 7 issues (1 high, 3 medium, 3 low) — **all fixed and re-verified** in the second commit (see its message). The high was an auto-mode restart loop; others covered a reconnect-hammer, a goroutine leak, a fail-toward-quit dialog, fractional settings input, and GPL/MPL license-text bundling.

## Hub wiring (commit 3) — works for CGNAT homes today

_(An earlier revision flagged "no prod hub" as a blocker. That's resolved — there **is** a live production hub, and the app is now wired to it and tested against it.)_

The app now ships with the production relay hub configured (`43.201.124.63:9443`, ap-northeast-2), so every install runs in **auto mode**: probe first, go **direct** when the home is reachable, **tunnel through the hub** when it isn't. That's what makes it useful to the IPv4/CGNAT majority, not just public-IPv6 homes.

**Hub TLS — pinning, not skip-verify.** Relay hubs run on bare IPs and self-sign, so CA verification can't trust them. Rather than an insecure skip-verify toggle (the MITM hazard the review flagged), the engine **pins the hub's exact leaf cert by SHA-256**: skip the chain check, require the presented leaf to match. MITM-proof without a CA. The pin applies only to the built-in hub; a user-supplied hub falls back to its own TLS trust. The same fingerprint mechanism will later be delivered over the broker's signed relay-list channel so the hub can rotate without a new release.

**Verified against the live prod hub:** the real engine attached in tunnel mode with the pin (connecting to the actual self-signed cert, no insecure flag), the hub allocated a public port and registered the relay with the **production broker**, and app bytes round-tripped hub → tunnel → engine → back. The hub accepts volunteers anonymously, so there's no token traveling over TLS.

## Remaining before a public release

- **Cost:** every CGNAT volunteer's full traffic transits the hub both ways (~$575/mo per sustained 20 Mbps on AWS egress per `deploy/relayhub/README.md`) — keep hubs on cheap-bandwidth hosts and scale horizontally.
- **Hub discovery (follow-up):** the hub IP is currently baked into the binary (chosen deliberately for speed). A follow-up moves hub address + fingerprint into the broker's signed relay-list so it rotates/fails-over without a new release.
- **Signing:** unsigned like the client today (macOS ad-hoc, Windows/Linux unsigned) — needs the Developer-ID / Windows identity for a friction-free install.
- **Abuse model:** shared-token / anonymous registration — no per-volunteer identity or revocation yet.


openrung-admin:
### Review dispositions (Codex)

Recording how the Codex review findings were resolved so the merge decision is clear.

**Fixed in this branch:**
- **Probe cert pinning** — the reachability probe ignored the hub's pinned cert, so auto mode always fell back to tunnel against the self-signed prod hub. Fixed by sharing one pin-aware TLS config between the tunnel dial and the probe; verified live against the prod hub (`166122b`).
- **Hub-outage recovery + no speculative direct** — auto mode now picks direct **only** when a probe positively confirms reachability (a public IPv6 doesn't imply inbound reachability), and the tunnel path re-probes to promote tunnel→direct when the hub confirms reachability. This removes a dead-relay hazard where a firewalled machine could be yanked into speculative direct on a hub blip and left advertising a dead endpoint (`4644002`).
- **Consent copy** — no longer implies an enforced bandwidth cap; states the speed setting is an advertised hint, not enforced yet.
- **Windows quit dialog** — sets `DefaultButton: "No"` on Windows (Wails' MB_YESNO only honors the safe default when it lowercases to `"no"`), so Enter keeps the relay running.
- **README** — documents the baked-in default hub, pinned cert, and auto/direct-only modes.
- Plus a new **Direct-only** connection mode so public-IP volunteers can run independent of the shared hub.

**Acknowledged, deferred (not a merge blocker), tracked for pre-wide-launch hardening:**
- **Direct relays aren't externally revalidated.** Direct endpoints are probe-verified at registration, but once a direct session is running only the local IPv6 *address* is watched (5-min recheck) — so a mid-session public-IPv4 change or a port-forward/firewall rule disappearing leaves the volunteer heartbeating a dead endpoint until broker ranking demotes it. Backstops (broker ranking + client multi-relay failover) bound the impact. Deferred because the clean fix is constrained — external revalidation needs the hub's dial-back probe, which needs the listen port that the observer already owns in direct mode — so it deserves a real design (ephemeral-port probe vs restart-to-revalidate vs broker-feedback) informed by usage data rather than a rushed pre-release patch. Should land **before a wide public launch** (a dead endpoint costs a censored user a connection attempt).

**Separately tracked (infra, not client):** the relay hub is currently a single instance — extended-outage resilience should come from a second hub rather than client-side guessing.

PR #50, opened 2026-07-11
desktop: multi-relay failover + connect probes + mid-session recovery

openrung-admin:
## What & why

The desktop client tried **exactly one** relay and declared "connected" if sing-box survived 400 ms — so a blocked top-of-list relay failed the whole connect, and a black-holing relay that passed no traffic still showed as connected. This ports the mobile clients' connect ladder to desktop and adds desktop-only mid-session recovery.

### Connect ladder (`internal/client`, `desktop/vpnservice/{connect_helpers,reachability,internetprobe,service}.go`)
- Walk the ordered usable candidate list. Per relay: TCP reachability probe (5 s, `relay_tcp_ms`) → start sing-box → wait for it to bind the inbound (real `tunnel_start_ms` via a readiness poll, replacing the fixed 400 ms grace) → verify **real internet through the tunnel** with a `generate_204` probe (12 s / 3 s / 500 ms) before promoting to CONNECTED.
- Failed relays are torn down and the next is tried; `relay_attempt_failed` per rung (1-based `attempt`), `relay_attempts` on `connection_succeeded`.
- Pinned `targetRelayID` stays pinned (no fallback); country targets stay in-country; broker ordering preserved (filter, don't re-sort). Targeted connects fetch `DirectoryRelayLimit=20` so the target is in the page.

### Mid-session recovery (`desktop/vpnservice/monitor.go`, desktop-only)
- A jittered through-tunnel health probe runs while connected. An unexpected sing-box exit, **or** 3 consecutive probe failures gated by a broker-front network-alive check (so a wifi blip / sleep doesn't churn relays), triggers **one** automatic re-ladder: fresh fetch, failed relay demoted to the end, status `connected → connecting → connected/failed`. User disconnect always wins; the OS proxy is released during the gap so traffic falls back to the normal network (availability over leak).

### Robustness / correctness (hardened across three adversarial review rounds)
- OS proxy is applied only once a candidate is proven, so a failing ladder never blackholes traffic; snapshot captured once, restored on exit.
- `Connect`/`Disconnect`/`Shutdown` serialized (`connectMu`) so overlapping connects can't orphan a live tunnel; status writes emit under one lock so a racing disconnect can never wedge the UI on a transient status.
- Telemetry manager gets a 15 s HTTP timeout so a hung flush can't gate recovery; `connection_succeeded` carries the full measurement set on every promote (including failover winners) so the broker ranking credits them.
- `internal/client/engine.go` gains a configurable `SingBoxRunner.KillGrace` (Windows has no `os.Interrupt`, so the ladder shortens the teardown grace).

## Scope / reviewer notes
- **No broker, mobile, or frontend contract changes.** `NativeVpnState` is unchanged; all recovery progress surfaces through existing log lines and status transitions.
- Branch is based on an older `main` tip; it doesn't touch any of the 4 newer main-only files, so the three-dot PR diff shows only this work.
- **17 new tests** cover the ladder, failover, probes, telemetry, and the specific concurrency races (including stress tests that reproduce the original bugs). All green under `-race`.
- **Verification caveat:** exercised via the full test suite with injected fakes for sing-box / broker / relays; a real-relay smoke test on a packaged build (needs live relays, the sing-box binary, and OS-proxy elevation) was not possible in this environment.

### DIFF
--- a/desktop/config/config.go
+++ b/desktop/config/config.go
@@ -49,7 +49,57 @@ const (
 	// request timeout. Must stay in sync with the mobile AppConfig's
 	// DISCOVERY_STAGGER_MS so every client races identically.
 	DiscoveryStagger = 2500 * time.Millisecond
+
+	// RelayTCPTimeout bounds the pre-connect TCP reachability check against a
+	// relay's public endpoint (it feeds relay_tcp_ms). Must stay in sync with
+	// the mobile RelayReachability.checkTcp timeout so both clients judge
+	// reachability identically.
+	RelayTCPTimeout = 5 * time.Second
+
+	// Internet probe: a connect is reported CONNECTED only after an end-to-end
+	// HTTP probe through the tunnel succeeds. Sweeps of InternetProbeURLs are
+	// retried every InternetProbeRetryDelay until InternetProbeOverallTimeout,
+	// each request bounded by InternetProbeRequestTimeout. Must stay in sync
+	// with the mobile InternetProbe constants.
+	InternetProbeOverallTimeout = 12 * time.Second
+	InternetProbeRequestTimeout = 3 * time.Second
+	InternetProbeRetryDelay     = 500 * time.Millisecond
+
+	// LadderKillGrace bounds how long a failed connect-ladder candidate's
+	// sing-box gets between interrupt and hard kill. Desktop-only: os.Interrupt
+	// is unsupported on Windows, so without this every failed candidate's
+	// teardown would cost the engine's full 5s default. Proxy mode holds no TUN
+	// device, so a hard kill is safe.
+	LadderKillGrace = 500 * time.Millisecond
+
+	// TunnelReadyTimeout bounds how long a candidate's sing-box has to bind its
+	// mixed inbound before the rung is judged failed (a config or bind error).
+	// It replaces a fixed post-launch grace, so a healthy engine that binds in
+	// tens of ms is not made to wait the whole window.
+	TunnelReadyTimeout = 5 * time.Second
+
+	// MaxRecoveryBackoff caps the Retry-After a rate-limited broker can impose on
+	// a mid-session recovery fetch, so a misbehaving or hostile front cannot
+	// suspend reconnection (and leave traffic on the normal network) for long.
+	MaxRecoveryBackoff = 60 * time.Second
+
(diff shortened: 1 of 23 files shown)


openrung-admin:
## Review of the follow-up commit (`6df7b06` "Fix desktop failover teardown and telemetry")

*(Posting as a comment, not an approval — GitHub won't let the PR author approve their own PR.)*

Verified by fast-forwarding my worktree to it, rebuilding, and running the core + desktop suites under `-race` (twice) — all green.

### Summary of the four changes

**1. Telemetry: measured `relay_failover` + broker-side handling.** This resolves a tradeoff I'd settled for in my own review rounds: emitting a second `connection_succeeded` per failover inflates dashboard success trends, but emitting a bare `relay_failover` instead meant the broker silently dropped the failover winner's ranking credit. This commit does it properly — the client emits one *measured* `relay_failover`, and the broker now counts it as a per-relay success observation (ranking + `relay_successes` in both store implementations) while excluding it from session-level attempt/success trends. Covered by a test asserting failover doesn't inflate trend successes.

**2. `MarkConnected` no longer resets `ConnectedAt` on a relay switch.** Real bug I missed — every failover was resetting the session's connected-duration clock, so `connection_duration_ms` on heartbeats/`connection_ended` only measured time since the last failover. Fixed with a test.

**3. Engine teardown hardening.** Process-group SIGINT/SIGKILL on Unix (catches helper children holding the port) plus a bounded 2s wait after hard-kill instead of blocking indefinitely on an unkillable process. Bounds worst-case candidate teardown, which my original `teardown()` didn't. The nil-on-cancel contract the supervisor relies on is preserved.

**4. Proxy-restore resilience.** `proxySet` is now marked before `Set()` (so a partially-applied Set still gets restored), a failed restore keeps the persisted snapshot and retries at cleanup/next launch instead of dropping it, and `Startup` only clears the crash-recovery snapshot once the restore actually succeeds. Consistent with the snapshot-retention invariant from my earlier fix round, with a new retry test.

### One operational note (non-blocking for this PR, but worth flagging before release)

The client now depends on the broker understanding `relay_failover` for failover winners to earn ranking credit. The prod broker doesn't auto-deploy (`broker-image.yml` is build-only), so it should be redeployed before/with the next desktop release that ships this — an old broker won't error on the unknown event, it'll just ignore it and failover winners get zero ranking signal until it's deployed.

I approve this revision.

PR #55, opened 2026-07-12
broker: authenticated node_class distinguishes foundation and volunteer relays

openrung-admin:
## What

Adds a broker-attested `node_class` (`foundation` | `volunteer`) so Foundation-operated and community-operated relays are cleanly distinguishable at the protocol level.

The class is operator provenance, not a quality score or trust bypass. Reliability and selection remain telemetry-driven, and every relay still goes through the same signature and transport verification.

## Design

- **Signed wire field:** `node_class` is optional on `RegisterRequest` (default `volunteer`) and always serialized on `Descriptor`. It is covered by the existing Ed25519 relay-list signature; older clients ignore the added JSON field.
- **Broker attestation:** a separate `OPENRUNG_FOUNDATION_TOKEN` authorizes Foundation claims. It must differ from the volunteer token. Missing or weaker credentials get `403`; the broker never silently downgrades a claim.
- **Self-contained relay credential:** a Foundation relay is configured by a single `OPENRUNG_FOUNDATION_TOKEN` (the same secret the broker accepts). Presenting it forces `node_class=foundation`, direct mode (`auto`/`tunnel` never run, so the bearer never reaches a hub probe), an HTTPS broker URL, and redirect refusal — for both registration and heartbeat. No separate `OPENRUNG_NODE_CLASS`/`OPENRUNG_MODE` is needed; an explicit conflicting class is a startup error. Routine volunteer and hub traffic continues to use the volunteer credential.
- **Client attestation check:** a Foundation relay refuses to serve mislabeled — `register()` verifies the broker echoed `node_class=foundation`, and any registration failure (an older broker that dropped the field, or a rejected claim) tears the relay down. Registration runs *after* xray and the public listener are up, so the broker never advertises a relay that cannot bind its port (a bind/xray failure aborts before any lease exists, so a restart loop can't perpetually refresh a dead row). The trade-off is a brief startup window before the current run re-attests, during which a cached client could reach the same operator-controlled machine — which then tears down if attestation is rejected. (An earlier "attest before opening the listener" ordering was reverted because it could strand a live Foundation lease with nothing serving; see the commit history.)
- **Endpoint ownership:** both memory and PostgreSQL stores enforce one current descriptor per `public_host:public_port`. A live Foundation endpoint cannot be seized by a volunteer credential; an authorized Foundation refresh replaces pre-squatted/old IDs, and an expired Foundation endpoint is reclaimable.
- **Lease guard:** heartbeating a Foundation descriptor requires the Foundation token. Refused heartbeats do not extend the lease, and PostgreSQL lookup failures propagate as service errors rather than false `404`s.
- **Freshness bounds:** an unattended Foundation row expires from the origin store after one lease TTL. That is not instant client revocation: API snapshots have a 30-minute `not_after`, the edge may serve a last healthy response for up to 15 minutes within that bound, and mirrors have a 24-hour bound. The desktop fallback cache now stops serving a snapshot after `not_after` plus the protocol's clock-skew allowance.

## Deployment safety

- Cloud bootstrap/user-data is not used for the privileged Foundation bearer.
- Broker Lightsail provisioning rejects `OPENRUNG_FOUNDATION_TOKEN`.
- Volunteer Lightsail and Hetzner helpers provision anonymous volunteer-class relays only and reject registration-token injection, including the Foundation token.
- Install Foundation credentials post-boot over an authenticated channel in a root-owned mode-`0600` env file, then recreate the container with `--env-file`; a restart does not reload changed environment variables.

## Verification

- `go test ./...`
- `go test -race ./internal/broker ./cmd/volunteer ./internal/volunteer`
- live PostgreSQL 16 broker suite via `OPENRUNG_TEST_POSTGRES_URL`
- `go vet` for the changed broker, relay, volunteer, and command packages
- desktop `go test ./vpnservice` and `go test -race ./vpnservice`
- Bash/POSIX shell syntax checks plus fail-closed cloud-helper and entrypoint-mode harnesses (the entrypoint mode/flag mapping is exercised against real busybox in `alpine:3.21`)
- `git diff --check`

## Deploy notes

- Add a distinct `OPENRUNG_FOUNDATION_TOKEN` to the broker's protected env file and recreate the broker container.
- Foundation relays set `OPENRUNG_FOUNDATION_TOKEN` (the same secret as the broker's), an HTTPS `OPENRUNG_BROKER_URL`, and `OPENRUNG_PUBLIC_HOST`. The token alone forces Foundation class, direct mode, HTTPS, and redirect refusal — no separate `OPENRUNG_NODE_CLASS` or `OPENRUNG_MODE`.
- **Infra prerequisite (not in this PR):** Foundation registration needs a TLS broker endpoint reachable from datacenter IPs. The plaintext origin used by community volunteers exists because Cloudflare challenges datacenter IPs, and is volunteer-only — so provisioning Foundation-class relays needs a CF IP-allowlist for their static IPs or direct TLS on the origin. Until that lands, this PR is safe to merge dormant: with no `OPENRUNG_FOUNDATION_TOKEN` set the mechanism is inert and every relay stays `volunteer`.
- If a broker is rolled back past `node_class`, follow the documented re-upgrade remediation in `deploy/broker/README.md`.


openrung-admin:
## Round 2 — addressing the registration-path bypass + rollout gaps

Thanks for the review. All points confirmed against the code and fixed in `3774bd5`.

**Blocker: registration-path authorization bypass.** Confirmed — the postgres upsert on `ON CONFLICT (public_host, public_port)` replaced the whole descriptor and set `node_class` from the request with no check on the existing row, so an anonymous volunteer could seize a foundation relay's entry and force a re-registration race. The heartbeat guard was on the wrong path.

Fix — a *live* foundation endpoint may only be overwritten by another foundation-class registration:
- Postgres: `DO UPDATE ... WHERE existing.node_class <> 'foundation' OR EXCLUDED.node_class = 'foundation' OR existing.expires_at <= registered_at`; a suppressed update → `403`. The `expires_at` disjunct keeps parity with the in-memory store's liveness check (an expired-but-unpruned foundation row stays reclaimable — caught by a follow-up adversarial pass, which otherwise had postgres over-blocking vs memory).
- Memory store: equivalent scan, so the guarantee holds on both backends.
- `EXCLUDED.node_class = 'foundation'` can only be true for a token-authorized caller, since the handler already gates the foundation claim.

**Credential path / cleartext.**
- The volunteer binary refuses `node_class=foundation` over a non-`https` broker URL (loopback excepted), before any network I/O.
- Tunnel/auto mode now **rejects** `node_class=foundation` (was warn-and-ignore): the class is unachievable through a hub, and proceeding would put the token in the hub Hello frame (cleartext when `-hub-tls=false`).

**Rollout wiring.**
- Broker `lightsail-up.sh` threads `OPENRUNG_FOUNDATION_TOKEN`.
- Volunteer `lightsail-up.sh` / `hetzner-up.sh` thread `OPENRUNG_NODE_CLASS` (default volunteer) and, on Lightsail, an optional token.
- **Called-out prerequisite:** foundation relays need a TLS broker endpoint. The plaintext origin (`54.238.185.205:8080`, used because Cloudflare challenges datacenter IPs) is now volunteer-only, so provisioning foundation-class relays needs either a CF IP-allowlist for their static IPs or direct TLS on the origin. That's infra, not code, and is documented rather than silently assumed.

**Verification.** Full `go test ./...` incl. the live-Postgres suite is green. End-to-end against a real broker + Postgres: anonymous overwrite → 403 with the foundation descriptor byte-for-byte intact, legit token refresh → 201, expired foundation endpoint reclaimable, foundation-over-http and tunnel-mode-foundation both fail closed. New tests cover each guard on both stores.

PR #58, opened 2026-07-13
broker: document end-to-end TLS for the CloudFront front origin leg

openrung-admin:
## What

The AWS CloudFront front (`E2PLKW8FO3JZSA` / `d2r7mdpyevvs1m.cloudfront.net`) reached the broker origin over **plaintext HTTP**, so the high-value **Foundation registration token** (`Authorization` on `POST /api/v1/volunteers/register`) crossed the CloudFront-edge → origin leg in cleartext. This PR is **docs only** — it records the infra change that closes that leg, reproducibly.

The leg is now encrypted end-to-end:

```
relay ──HTTPS──► CloudFront edge ──HTTPS(:443)──► Caddy ──HTTP(127.0.0.1:8080)──► broker
```

## Infra applied (on the broker box + AWS API — not code in this PR)

- **Caddy** on `typhoon-broker` terminates a Let's Encrypt cert (Caddy-default **ECDSA P-256**) on `:443` and reverse-proxies to the broker on `:8080`. Additive: the broker container and the plaintext `:8080` volunteer path are **untouched**; the cert auto-renews natively.
- **CloudFront** origin flipped `http-only` → `https-only`, and origin request policy `Managed-AllViewer` → `Managed-AllViewerExceptHostHeader`.

## The 502 root cause (single): origin SNI, not cipher/key-type

With `Managed-AllViewer`, CloudFront forwards the viewer `Host` header and **uses it as the origin SNI** (`d2r7mdpyevvs1m.cloudfront.net`), for which Caddy has no cert → handshake rejected → 502. `Managed-AllViewerExceptHostHeader` makes CloudFront send the **origin** hostname as Host+SNI (still forwarding `Authorization`, cookies, query strings), which Caddy serves and routes.

> An earlier revision wrongly blamed "RSA-only origin ciphers." CloudFront's captured origin `ClientHello` offers `ECDHE-ECDSA-*`, P-256, `ecdsa_secp256r1_sha256`, and TLS 1.3 — verified end-to-end that the **default ECDSA** cert works. `key_type rsa2048` was dropped.

## Verified

- Origin direct HTTPS → **200**, publicly-trusted cert.
- Discovery through CloudFront `GET /api/v1/relays` → **200**, signed, from a datacenter IP.
- **Foundation-token forwarding (the load-bearing property) — with a *discriminating* test.** `POST {}` → 400 would **not** prove it (prod allows anonymous registration, so an unauthenticated request gets the same 400). Instead: `POST {"node_class":"foundation"}` with `public_host` omitted (no relay created) — **with** the token → `400 public_host is required` (token survived the origin leg); **without** → `403 requires the foundation registration token`. `400` vs `403` is the signal. The runbook's recipe keeps the token out of shell history and process argv (`curl -H @tmpfile`, `shred`). Cross-check: Caddy access log shows `Authorization: ["REDACTED"]` (forwarded; value never written to disk).
- Volunteer plaintext `:8080` and the Cloudflare Worker front both **unaffected**.

## Files

- `deploy/broker/origin-tls.md` — full runbook (setup, cert/renewal, firewall, CloudFront settings, the SNI root cause, token-safe verification, rollback, follow-ups).
- `deploy/broker/Caddyfile` — deployed config (byte-for-byte matches the box; default ECDSA; comments reflect live state).
- `deploy/broker/README.md`, `deploy/broker-proxy/README.md` — pointers + scope the Worker's plaintext-origin note.

## Follow-ups

- **Loopback-wide rate-limit / telemetry collapse — resolved; now keyed per-CloudFront-edge.** `OPENRUNG_TRUSTED_PROXY_CIDRS=127.0.0.1/32,::1/128` applied to the broker restores pre-proxy per-edge keying (unspoofable CloudFront edge IP Caddy forwards).
- **Per-*viewer* rate-limit / telemetry — OPEN (achievable, not yet done).** Per-edge is a current *implementation* choice, not a CloudFront limitation: clients sharing an edge still share its 2 req/s / burst-30 bucket + telemetry identity (can saturate under mass failover). `AllViewerExceptHostHeader` already forwards **`CloudFront-Viewer-Address`** (viewer ip:port), verified **unspoofable through CloudFront** (a forged value is overwritten with the real viewer IP). Adopting it needs the origin to first authenticate the request came via CloudFront (shared-secret origin header, or lock `:443` to CloudFront origin-facing ranges), since `:443` is internet-facing and a direct hit could forge the header.
- **Cloudflare Worker front origin leg still plaintext `:8080`** — could point at `https://broker-origin.openrung.org` now; separate change to the higher-traffic primary front.

### DIFF
--- a/deploy/broker-proxy/README.md
+++ b/deploy/broker-proxy/README.md
@@ -93,11 +93,15 @@ Logs: `wrangler tail openrung-broker-proxy`.
 
 ## Known limitations / follow-ups
 
-- **Origin leg is plaintext HTTP.** The censorship-relevant leg (client → Cloudflare) is
-  encrypted, but Cloudflare → origin is HTTP over the public internet. Harden by giving the origin
-  a Cloudflare Origin CA cert and switching SSL/TLS mode to Full (strict), by fronting via a
+- **Origin leg is plaintext HTTP (this Worker front only).** The censorship-relevant leg
+  (client → Cloudflare) is encrypted, but *this Worker's* Cloudflare → origin subrequest is still
+  HTTP (`http://broker-origin.openrung.org:8080`) over the public internet. Harden by giving the
+  origin a Cloudflare Origin CA cert and switching SSL/TLS mode to Full (strict), by fronting via a
   Cloudflare Tunnel (no public origin port), or by firewalling the origin to Cloudflare egress
-  ranges only.
+  ranges only. **The independent AWS CloudFront front already has end-to-end TLS to the origin**
+  (via a Caddy Let's Encrypt terminator on `:443` — see `deploy/broker/origin-tls.md`); the same
+  `:443` origin endpoint is now available to point this Worker at
+  (`https://broker-origin.openrung.org`) to close its leg too.
 - **Client IP recovery (done).** The Worker forwards the real client IP as `X-Forwarded-For`, and
   the broker now honors `CF-Connecting-IP` / `X-Forwarded-For` **only when the request arrives from a
   trusted proxy** (Cloudflare's published ranges by default; extend via `OPENRUNG_TRUSTED_PROXY_CIDRS`).
--- a/deploy/broker/Caddyfile
+++ b/deploy/broker/Caddyfile
@@ -7,13 +7,13 @@
 # for community volunteer relay registration and for the Cloudflare Worker front
 # (broker.openrung.org), neither of which pass through here.
 #
-# Serves its cert on SNI broker-origin.openrung.org. AWS CloudFront (distribution
-# E2PLKW8FO3JZSA) must be configured to send THAT hostname as the origin SNI —
-# i.e. the behavior must use the AllViewerExceptHostHeader origin-request policy,
-# NOT AllViewer (which forwards the viewer Host and makes CloudFront use the
-# distribution domain as SNI, for which this proxy has no cert). See
-# origin-tls.md. The default ECDSA (P-256) leaf is fine for CloudFront and every
-# other client.
+# Serves its cert on SNI broker-origin.openrung.org. CloudFront (distribution
+# E2PLKW8FO3JZSA) is configured to send THAT hostname as the origin SNI via the
+# AllViewerExceptHostHeader origin-request policy. (Do NOT switch it back to
+# AllViewer: that forwards the viewer Host, which CloudFront then uses as the
+# origin SNI — the distribution domain, which this proxy has no cert for, so the
+# handshake 502s. See origin-tls.md.) The default ECDSA (P-256) leaf is fine for
+# CloudFront and every other client.
 
 {
 	# No ACME account email on purpose: Caddy auto-renews, so expiry notices are
@@ -26,13 +26,13 @@ broker-origin.openrung.org {
 		# Client-IP provenance for the CloudFront path.
 		#
 		# The broker honors forwarded client IPs only from trusted proxies
-		# (internal/broker/clientip.go). CloudFront's behavior forwards ALL viewer
-		# headers except Host, so a client could inject X-Forwarded-For /
-		# CF-Connecting-IP. Strip both and set XFF to our own immediate peer — the
-		# UNSPOOFABLE CloudFront edge IP — so that when the broker trusts this
-		# loopback hop (OPENRUNG_TRUSTED_PROXY_CIDRS=127.0.0.1/32,::1/128) it
-		# buckets CloudFront-fronted requests by the CloudFront edge IP rather than
-		# collapsing them all onto 127.0.0.1.
+		# (internal/broker/clientip.go) and trusts this loopback hop:
+		# OPENRUNG_TRUSTED_PROXY_CIDRS=127.0.0.1/32,::1/128 IS set on the broker.
+		# CloudFront forwards ALL viewer headers except Host, so a client could
+		# inject X-Forwarded-For / CF-Connecting-IP. Strip both and set XFF to our
+		# own immediate peer — the UNSPOOFABLE CloudFront edge IP — so the broker
+		# buckets CloudFront-fronted requests (rate limit, telemetry) by that edge
+		# IP instead of collapsing them all onto 127.0.0.1.
 		header_up -CF-Connecting-IP
 		header_up X-Forwarded-For {remote_host}
 	}
--- a/deploy/broker/Caddyfile
+++ b/deploy/broker/Caddyfile
@@ -7,39 +7,32 @@
 # for community volunteer relay registration and for the Cloudflare Worker front
 # (broker.openrung.org), neither of which pass through here.
 #
-# Serves its cert on SNI broker-origin.openrung.org, which is exactly the SNI
-# AWS CloudFront (distribution E2PLKW8FO3JZSA) sends to its origin.
+# Serves its cert on SNI broker-origin.openrung.org. AWS CloudFront (distribution
+# E2PLKW8FO3JZSA) must be configured to send THAT hostname as the origin SNI —
+# i.e. the behavior must use the AllViewerExceptHostHeader origin-request policy,
+# NOT AllViewer (which forwards the viewer Host and makes CloudFront use the
+# distribution domain as SNI, for which this proxy has no cert). See
+# origin-tls.md. The default ECDSA (P-256) leaf is fine for CloudFront and every
+# other client.
 
 {
 	# No ACME account email on purpose: Caddy auto-renews, so expiry notices are
 	# non-essential, and we avoid tying an operator identity to the issuance.
 	# Let's Encrypt production CA + ToS are accepted implicitly by Caddy.
-
-	# RSA (not Caddy's default ECDSA) leaf key. AWS CloudFront's origin-facing TLS
-	# cipher list is RSA-only: it offers the origin ECDHE-RSA / RSA suites and NO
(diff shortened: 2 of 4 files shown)


openrung-admin:
You're right — thanks for catching this. The "RSA-only" diagnosis was wrong, and I've fixed it in `c5b0553`.

**What actually happened:** I decoded the CloudFront origin `ClientHello` that Caddy logged during the outage. It offers ECDSA support outright:

- **Cipher suites:** `ECDHE-ECDSA-AES128-GCM-SHA256`, `ECDHE-ECDSA-AES256-GCM-SHA384` (+ CBC variants), plus the TLS 1.3 suites (which are signature-agnostic).
- **Curves:** `secp256r1 (P-256)`, `x25519`, `secp384r1 (P-384)`.
- **Signature schemes:** `ecdsa_secp256r1_sha256`, `ecdsa_secp384r1_sha384`, `ecdsa_secp521r1_sha512`.
- **Versions:** TLS 1.3, TLS 1.2.

So it matches AWS's documented origin ciphers, ECDSA included. My earlier "confirmation" was a bad test — I forced `openssl` to offer *only* `ECDHE-RSA` ciphers against the ECDSA cert, which trivially can't complete and says nothing about what CloudFront offers.

**The real and only root cause** was the SNI: `Managed-AllViewer` forwards the viewer `Host` and CloudFront uses it as the origin SNI (`d2r7mdpyevvs1m.cloudfront.net`), for which Caddy had no cert → `no certificate available` → 502. `AllViewerExceptHostHeader` fixes it.

**Verified on prod:** I reverted the live origin to Caddy's default **ECDSA P-256** cert (dropped `key_type rsa2048`) and CloudFront still returns `200` from both a residential and a datacenter IP, with `POST …/register {}` → `400`. The runbook and Caddyfile now document SNI as the single cause, with an explicit note that ECDSA works and RSA is not needed.

openrung-admin:
Fixed in `3807d1d` — you're right, that was a false positive.

I replaced the check with your discriminating test and ran it against prod (token kept server-side, never echoed; `public_host` omitted so no relay is created):

| request through CloudFront | result |
|---|---|
| **A.** `{"node_class":"foundation"}` **with** real Foundation token | **`400 {"error":"public_host is required"}`** |
| **B.** `{"node_class":"foundation"}` **without** token | `403 {"error":"node_class foundation requires the foundation registration token"}` |
| **C.** `{"node_class":"foundation"}` with a **wrong** token | `403` (same as B) |

`400` vs `403` is the signal: if `AllViewerExceptHostHeader` (or the origin leg) had dropped `Authorization`, case A would have returned `403` like B/C. It returned `400`, so the token reached the broker intact.

No-secret cross-check: the Caddy access log records `request.headers.Authorization = ["REDACTED"]` on a CloudFront-fronted request that carried one — confirming forwarding, and confirming Caddy redacts the value so the token is never written to disk. The runbook's Verify section now documents the discriminating test and calls out explicitly why `POST {}` → 400 does **not** prove header forwarding under anonymous registration.

openrung-admin:
Fixed in `e530c6b`. Both stale comments now match the live/resolved state:

- **Loopback trust** — was phrased as a future "when the broker trusts this loopback hop"; now present tense: `OPENRUNG_TRUSTED_PROXY_CIDRS=127.0.0.1/32,::1/128 IS set on the broker`, so the forwarded CloudFront edge IP is actually what's used for rate-limit/telemetry bucketing.
- **Policy name** — the header comment now states CloudFront **is configured with** `AllViewerExceptHostHeader`, and names `AllViewer` only inside an explicit "do NOT switch back" warning (with the 502 rationale), rather than as prescriptive text that could read as current state.

Comments-only change; I re-synced it to the box and confirmed the committed `Caddyfile` still matches `/etc/caddy/Caddyfile` byte-for-byte (origin + CloudFront both still 200 after the reload).

openrung-admin:
Fixed in `c35f5dc`. The recipe no longer routes the token through history or argv.

New approach (verified live — still `400` with token / `403` without):
- Token goes **file → file** via `awk` from the root-owned `broker.env` into a **tmpfs** header file (`/dev/shm/ft.hdr`, `umask 077`) — it's never an argv, never echoed, never in history.
- curl reads it with **`-H @/dev/shm/ft.hdr`** (not `-H "…$FT"`), so it stays out of `ps`/argv — addressing curl's own warning that command-line args aren't reliably hidden.
- **`shred -u`** cleanup afterward (and it's RAM-backed, never on disk).
- For off-box use I documented the hidden-prompt variant: `read -rs FT; printf 'Authorization: Bearer %s\n' "$FT" > /dev/shm/ft.hdr; unset FT` — `read`/`printf` are builtins, so no process argv either.

Dropped the "leading space / token in a var" guidance entirely, per your note.

openrung-admin:
Fixed in `6d63268` — you're right, "per-client resolved" overclaimed. Corrected framing:

- What's resolved is the **loopback-wide collapse** (all CloudFront traffic → `127.0.0.1`, one bucket). Adding `OPENRUNG_TRUSTED_PROXY_CIDRS=127.0.0.1/32,::1/128` restores **per-edge** keying — i.e. exactly the pre-proxy behavior, keyed by the CloudFront **edge** IP.
- It is **not per-viewer**, and the runbook now says so explicitly: clients behind one edge still share that edge's 2 req/s / burst-30 relay-list bucket and telemetry identity, and **a mass-failover surge** — many clients in a region funneling through a handful of nearby edges — can still push a per-edge bucket to its limit.
- I also documented *why* per-edge is deliberate rather than a shortfall: the only per-viewer signal CloudFront gives the origin is the client-*appended* `X-Forwarded-For`, which a client can forge to evade the limiter or poison telemetry, so the unspoofable edge IP is the correct key. True per-viewer limiting would need CloudFront to attest the viewer IP in a header the origin can trust, which it doesn't.

Heading changed to "Loopback-wide rate-limit / telemetry collapse — RESOLVED; residual per-edge aggregation is by design."

openrung-admin:
Updated the PR description in sync with the runbook. The "Verified" section no longer claims `POST {}` → 400 proves the token path — it now documents the discriminating `node_class:foundation` test (400 with token / 403 without), the token-safe recipe, and the Caddy-log cross-check. Also refreshed the body's root-cause (SNI, ECDSA) and the per-edge (not per-client) follow-up framing to match the corrected commits.

openrung-admin:
You're right — corrected in `233093d`, and I reproduced your spoof test to confirm before rewriting.

Sent through the distribution with a **forged** header:
```
curl -H 'CloudFront-Viewer-Address: 203.0.113.99:4444' 'https://d2r7mdpyevvs1m.cloudfront.net/api/v1/relays?...'
```
The origin (Caddy access log) received:
```
Cloudfront-Viewer-Address = ['159.117.71.211:59399']   # my real viewer IP, not the forged value
Cloudfront-Viewer-Asn/City/Country/Latitude/Longitude/Tls = [...]   # full suite also forwarded
```
So `AllViewerExceptHostHeader` **does** deliver `CloudFront-Viewer-Address` and CloudFront **overwrites** any client-supplied value — it's attested/unspoofable through the distribution.

The runbook now:
- reframes per-edge as the **current implementation**, not a CloudFront limitation;
- adds **per-viewer as an OPEN follow-up**: map the IP part of `CloudFront-Viewer-Address` into the broker's client-IP key;
- flags the caveat I don't want a future implementer to miss — `:443` is internet-facing, so a **direct** (non-CloudFront) hit could forge `CloudFront-Viewer-Address`; it must only be trusted after authenticating the request came via CloudFront (shared-secret origin header, or lock `:443` ingress to CloudFront's origin-facing ranges), else per-viewer keying would be *more* spoofable than the edge-IP key, not less.

PR body's follow-up section updated to match.

openrung-admin:
Fixed in `3e36c86`. Replaced `awk -F= '…{print $2}'` with a first-prefix-only strip:

```sh
sed -n 's|^OPENRUNG_FOUNDATION_TOKEN=|Authorization: Bearer |p' /etc/openrung/broker.env
```

`sed` removes only the leading `OPENRUNG_FOUNDATION_TOKEN=` and emits the rest of the line verbatim, so any `=` in the value survives. Verified on the box:
- `OPENRUNG_FOUNDATION_TOKEN=abc==` → `Authorization: Bearer abc==` (the old `awk -F=` gave `abc`).
- Re-ran the live discriminating test with the sed extraction: **400 with token / 403 without** — unchanged.

(The prod token is 64-char hex, so my earlier runs weren't affected — but you're right that the recipe must not assume that, since the broker doesn't constrain the token to hex.)

openrung-admin:
Fixed in `d7d5066`.

- **Predictable path → `mktemp`.** No more fixed `/dev/shm/ft.hdr`. `mktemp` gives a root-owned, `0600`, unpredictable name — nothing to pre-create or symlink in a shared dir, and it sidesteps the "`umask` can't fix an existing file" problem entirely (the file is freshly created).
- **Cleanup is now actually always.** `trap "shred -u … || rm -f …" EXIT` (and `INT TERM` in the off-box variant) removes the file however the shell exits — success, error, or signal — instead of relying on a final line that a mid-script failure would skip. Verified on the box: file is `0600`/root, and the EXIT trap fires and removes it even on an error exit (`set -e; false`).
- **Portability.** The off-box recipe is forced through `bash -c` (consistent trap behavior), uses a hidden `read` prompt, and is explicitly labeled that **macOS has no `/dev/shm` and no GNU `shred`**, so it uses `mktemp` + `rm`.

Live re-run of the exact documented on-box block: `with-token: 400 / no-token: 403`, unchanged.

PR #72, opened 2026-07-15
Reimplement foundation-up.sh with security fixes (redo of #68)

openrung-admin:
## Why

#68 added `foundation-up.sh` but was merged without the review sequence and reverted in #71. This PR reimplements the same tooling — `create` / `convert` / `update` for Foundation relays, wrapping (never modifying) `lightsail-up.sh` — with the issues from a security audit of #68 fixed. Since #68 carried no review record, the audit findings are documented here in full.

### 1. The Foundation bearer was routed through a third-party CDN by default

#68 defaulted `OPENRUNG_BROKER_URL` to `https://d2r7mdpyevvs1m.cloudfront.net`. Every registration and heartbeat would carry `OPENRUNG_FOUNDATION_TOKEN` through CloudFront, where viewer TLS terminates — the bearer is decrypted at every edge POP, and its exposure depends on distribution config staying correct (an `http-only`-origin or `Managed-AllViewer` regression would silently reopen the cleartext origin leg that `deploy/broker/origin-tls.md` was written to close). It also collapses broker-side per-IP rate limiting onto shared CloudFront edge IPs (the broker keys the CloudFront path per-edge, per origin-tls.md), so a fleet of relays near one POP shares one 2 req/s bucket — a heartbeat-starvation risk with 13+ relays mostly in one region.

**Fix:** default to `https://broker-origin.openrung.org` — the broker's own TLS origin (Caddy, publicly-trusted LE cert, set up 2026-07-13). The token is only ever decrypted on the broker box, and the broker rate-limits each relay by its real IP (Caddy forwards `X-Forwarded-For={remote_host}`, trusted via the loopback proxy CIDRs). Fronts exist for censored *clients*; datacenter relays have no reason to traverse one. Note for rollout: existing fleet env files keep whatever URL they were converted with until `convert` is re-run — `update` deliberately does not rewrite credentials.

### 2. `eval` on an environment variable, and token-handling hardening

#68 ran `eval "$OPENRUNG_FOUNDATION_TOKEN_CMD"` — full shell evaluation inside the script's own context, from a process that holds fleet SSH keys and AWS credentials. **Fix:** the token command runs via `bash -c` in a child shell; only its first output line is taken (matching `pass show` semantics). Additionally: xtrace is forced off so `bash -x` can never trace the token (test-covered), the token is charset-validated before being written (an embedded newline would inject extra variables into the env file), and the remote env-file write is atomic (`tmp` + `mv`) so a dropped connection can't leave a truncated credential file. Unchanged from #68: the token is never in argv on either side and travels only over SSH stdin.

### 3. Unvalidated remote-derived values flowed into privileged commands and the operator's terminal

#68 interpolated whatever a host returned — container name, env-file path, `OPENRUNG_PUBLIC_HOST`, `OPENRUNG_LABEL`, raw `docker logs` lines — into subsequent `sudo` command strings (including the `sh -c 'cat > ${target}'` that writes the token file) and printed them raw to the controlling terminal. A compromised relay could inject shell into the commands the script runs against it, or ANSI escape sequences into the operator's terminal. **Fix:** every remote-read value must pass a strict allowlist before reuse — container name must be exactly `openrung-relay`/`openrung-volunteer`, env-file path exactly the canonical/legacy path, identity values must match tight regexes — and display-only strings (log lines) are stripped of control characters. Operator-set values that reach remote commands (`OPENRUNG_IMAGE`, `OPENRUNG_ENV_FILE`, hosts) are validated too.

### 4. `update` verified nothing Foundation-specific

The "self-contained verification" (running + registered ⇒ broker attested `foundation`) is sound **only when the relay presents the token** — that's what forces the class claim the relay then cross-checks (`cmd/relay` exits on a mismatched attestation; the broker 403s an unauthorized `foundation` claim). #68's `update` never checked the env file it reused, so a host whose env file lacked the token would register as a volunteer and still "verify". **Fix:** `update` requires an `OPENRUNG_FOUNDATION_TOKEN=` line in the env file (presence check only; the value is never read back), making the verification's precondition hold on every path.

### 5. Trust-on-first-use at the exact moment the credential is shipped

#68 used `StrictHostKeyChecking=accept-new` and immediately streamed the token over that first connection — a first-contact MITM gets the fleet credential. **Fix:** before the first connection that will carry the token, the host's SSH keys are fetched over the authenticated Lightsail API (`get-instance-access-details`, with retries on `create` while a new instance settles; reverse lookup by IP for `convert`/`update`) and pinned into `known_hosts`. Non-Lightsail hosts fall back to `accept-new` with an explicit warning. `accept-new` still never accepts a *changed* key.

### 6. `create` could never have worked with one of its two documented token sources

`lightsail-up.sh` refuses to run when `OPENRUNG_FOUNDATION_TOKEN` is set, even empty (its `${VAR+x}` user-data guard, `lightsail-up.sh:29`) — so #68's `create` failed immediately whenever the token came via the plain variable rather than `_CMD`. Evidence the path was never executed end-to-end. **Fix:** `create` invokes the provisioning child with `env -u` stripping `OPENRUNG_FOUNDATION_TOKEN`, `OPENRUNG_FOUNDATION_TOKEN_CMD`, `OPENRUNG_VOLUNTEER_TOKEN`, and `OPENRUNG_NODE_CLASS`. This both unbreaks the flow and makes the no-token-in-user-data invariant **structural**: the provisioning child never holds the credential, so no bug in it could ever embed one. (Covered by a stubbed-child test that enforces the real guard.)

### 7. Smaller hardening and correctness items

- **Single token copy, canonical path:** `convert` writes `/etc/openrung/relay.env` and, once the relay verifies, removes a legacy `volunteer.env` (`shred -u || rm`). #68 instead overwrote the legacy path in place forever. The rollback container doesn't need the file (env is baked at create).
- **Failure paths print an in-place rollback command** (the previous container is kept, stopped, as `openrung-relay-old`), instead of only "check the logs".
- `usage()` no longer `sed`-extracts fixed line numbers from its own header (silent breakage when the header moves); verification window widened to 60s; live output from `lightsail-up.sh` streams during `create` instead of appearing only at the end.

## What #68 got right and is preserved

Wrapper-not-replacement of `lightsail-up.sh`; token only over SSH stdin, post-boot, never user-data; root-owned mode-0600 env file; tokenless `update` for routine image rolls; sequential fail-fast fleet updates; identity reuse on `convert` (no silent relabel/re-home); legacy `openrung-volunteer`/`volunteer.env` detection; HTTPS-broker fail-fast; container hardening parity with `lightsail-up.sh` including the deliberate absence of `no-new-privileges` (it would break the `cap_net_bind_service` 443 bind); the attestation-based self-contained verification (now with its precondition enforced).

## Testing

- `bash -n` clean; `shellcheck` (stable, via Docker) clean — 3 informational SC2029 notes only, all intentional client-side expansion of allowlist-validated values.
- 17 offline tests, all passing: usage/dispatch; HTTPS guard; token-source missing/empty/whitespace/failing-command; multi-line token command (first line taken, proceeds past token stage); allowlist rejections for hostile `OPENRUNG_IMAGE`, `OPENRUNG_ENV_FILE`, and host arguments; `create` fails before provisioning on bad broker URL or missing token; **`bash -x` trace-leak test** (a known token value never appears in trace output); and a **stubbed `lightsail-up.sh` `create` flow** proving the `env -u` guard, handoff-line parse, host-key-pin fallback warning, and wait-loop entry.
- **Not run against production hosts** — deploy authority for this repo requires the operator in the loop, and even read-only SSH was skipped. Suggested first live steps: `update` against one relay (no token involved), then `convert` against one relay to move it to the direct-origin broker URL, then the fleet.

## Review notes

- Same soft coupling as #68: `create` parses the `OPENRUNG_RELAY name=… ip=…` handoff line from `lightsail-up.sh:125`; it fails loudly if that line is reformatted.
- The direct-origin default depends on the origin-tls.md invariants: `broker-origin.openrung.org` stays DNS-only and Caddy stays up on the broker box (`:443`). Volunteer provisioning on plaintext `:8080` is unaffected.
- CI has no shell linting; shellcheck for `deploy/**/*.sh` would be a cheap follow-up.

### DIFF
--- a/.github/workflows/go-checks.yml
+++ b/.github/workflows/go-checks.yml
@@ -16,6 +16,17 @@ permissions:
   contents: read
 
 jobs:
+  relay-deploy-script:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Foundation deploy state matrix
+        run: |
+          bash -n deploy/relay/foundation-up.sh
+          bash -n deploy/relay/foundation-up_test.sh
+          bash deploy/relay/foundation-up_test.sh
+
   go:
     runs-on: ubuntu-latest
     steps:
--- a/deploy/relay/README.md
+++ b/deploy/relay/README.md
@@ -87,6 +87,70 @@ also set `OPENRUNG_NODE_CLASS` or `OPENRUNG_MODE`:
 traverses. A Foundation-operated relay hub therefore does not make the
 volunteer-run relays tunneled through it Foundation-operated.
 
+#### Automating the post-boot step
+
+[`foundation-up.sh`](foundation-up.sh) automates the two-step dance above. It
+**wraps** `lightsail-up.sh` rather than changing it: the token still reaches the
+host only over SSH, after boot, and never through user-data — `create` even
+strips the token variables from the provisioning child's environment, so no bug
+in `lightsail-up.sh` could ever embed the credential.
+
+```sh
+# Provision a new Lightsail host and install credentials on it.
+OPENRUNG_FOUNDATION_TOKEN_CMD='pass show openrung/foundation-token' \
+  deploy/relay/foundation-up.sh create
+
+# Promote hosts that are already serving as volunteer-class relays.
+OPENRUNG_FOUNDATION_TOKEN_CMD='pass show openrung/foundation-token' \
+  deploy/relay/foundation-up.sh convert 203.0.113.10 203.0.113.11
+
+# Roll a new image across the fleet. Needs no token: the credentials already on
+# each host are reused as-is.
+OPENRUNG_IMAGE=ghcr.io/openrung/openrung-relay:sha-abc1234 \
+  deploy/relay/foundation-up.sh update 203.0.113.10 203.0.113.11
+```
+
+| Variable | Default | Meaning |
+|----------|---------|---------|
+| `OPENRUNG_FOUNDATION_TOKEN_CMD` | — | Command printing the token (`pass show …`, `aws secretsmanager get-secret-value …`, `op read …`). Preferred. Run via `bash -c`; the first output line is the token. |
+| `OPENRUNG_FOUNDATION_TOKEN` | — | The token itself. Fallback for CI; prefer the command form so the secret is not sitting in your environment. |
+| `OPENRUNG_IMAGE` | `…/openrung-relay:main` | Image to run. Pin a `sha-…` tag (or an `@sha256:…` digest) for reproducible rolls. |
+| `OPENRUNG_BROKER_URL` | `https://broker-origin.openrung.org` | The broker's **direct TLS origin** (see `deploy/broker/origin-tls.md`), not a CDN front: a front would decrypt the Foundation bearer at every edge POP and collapse per-relay rate limiting onto shared edge IPs. Must be HTTPS; the script fails fast otherwise. |
+| `OPENRUNG_ENV_FILE` | `/etc/openrung/relay.env` | Where credentials live. `convert` writes the canonical path and, once the relay verifies, removes a legacy `volunteer.env` so exactly one copy of the token remains on disk. |
+| `OPENRUNG_SSH_KEY` / `OPENRUNG_SSH_USER` | `~/.ssh/id_ed25519_openrung` / `ubuntu` | SSH access to the host. |
+| `OPENRUNG_REGION` | `ap-northeast-1` | Region used to look up Lightsail-published SSH host keys for pinning. |
+
+Notes on the design:
+
+- **The token is never a command-line argument.** `argv` is world-readable via
+  `/proc` and is retained in shell history, so it is read from the environment or
+  a command (run via `bash -c`, never `eval`) and streamed to the host over
+  stdin. The script keeps shell tracing (`bash -x`) disabled so the token cannot
+  leak into trace output, and the env-file write is atomic (tmp + rename).
+- **SSH host keys are pinned out-of-band when possible.** Before the first
+  connection that will carry the token, the host's SSH keys are fetched over the
+  authenticated Lightsail API (`get-instance-access-details`) and written to
+  `known_hosts`, upgrading first contact from trust-on-first-use to verified.
+  Non-Lightsail hosts fall back to `accept-new` with a warning.
+- **Everything read back from a host is allowlist-validated** (container names,
+  env-file paths, identity values) before it is reused in a privileged command
+  or echoed to the terminal, so a compromised relay cannot inject shell or
+  terminal escapes into the operator's session.
+- **`update` needs no token**, because it reuses the env file the host already
+  has — but it refuses to run against an env file that lacks
+  `OPENRUNG_FOUNDATION_TOKEN` (presence check only), so it cannot silently
+  "verify" a non-Foundation host.
+- **`update` is sequential and fails fast.** A bad image stops the roll at the
+  first host rather than taking down every relay, and failures print an in-place
+  rollback command (the previous container is kept, stopped, as
+  `openrung-relay-old`).
+- **Verification is self-contained.** A relay that presents the Foundation token
+  forces `node_class=foundation` and exits during startup if the broker attests
+  any other class, so a container that is still running and has logged a
+  registration *is* the proof — no broker query needed.
+- Legacy `openrung-volunteer` container and `volunteer.env` names are detected,
+  so hosts predating the rename are handled without manual cleanup.
+
 ### Stable relay identity (recommended)
 
 Without an explicit identity, the relay generates a fresh one on every restart.
--- a/deploy/relay/README.md
+++ b/deploy/relay/README.md
@@ -164,9 +164,10 @@ Notes on the design:
   kept, stopped, as `openrung-relay-old`) is only printed once the old container
   has actually been swapped out — never on a pull failure, where the live relay
   is untouched and "rolling back" would destroy it. The swap also refuses to
-  discard an existing `-old` backup while the "live" container is not running
-  (the signature of a previously failed roll), so re-running after a failure
-  cannot delete the last healthy relay.
+  discard an existing `-old` backup unless the current container has itself
+  registered and stayed up — a failed rollout can be running yet never
+  register — so re-running after a failure can never delete the last
+  known-good relay.
 - **Verification is self-contained.** A relay that presents the Foundation token
   forces `node_class=foundation` and exits during startup if the broker attests
   any other class, so a container that logged a registration and is running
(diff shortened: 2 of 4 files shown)


openrung-admin:
Codex review addressed in 1a41056 — all four findings confirmed real and fixed, plus a round of adversarial verification over the fixes themselves.

## The four findings

**1. `convert` destroyed existing relay configuration (P1) — fixed.** `convert` no longer writes a fixed five-variable file. It stages the host's existing `OPENRUNG_*` settings *on the host itself* — canonical env file + legacy `volunteer.env` (both kept when present; docker's last-occurrence-wins gives the canonical file per-key precedence while legacy-only keys survive), else the running container's env — and rewrites only `OPENRUNG_BROKER_URL` and `OPENRUNG_FOUNDATION_TOKEN`. A pinned identity (Reality keys, client id, short id), capacity limits, and camouflage all survive conversion, and nothing preserved ever transits the connection. Two related fixes that fell out of verifying this one: the `OPENRUNG_LISTEN_HOST=0.0.0.0` default is **gone** (the binary's own default is `::` dual-stack, so injecting 0.0.0.0 would have silently cut IPv6 on any host relying on the default — the bootstrap scripts pass it explicitly so the fleet is unaffected either way), and image-internal `OPENRUNG_XRAY_PATH`/`OPENRUNG_CONFIG_OUT` are dropped from preservation so a `docker inspect` fallback can't pin image ENV onto the host. One honest caveat now documented instead of overclaimed: bootstrap-provisioned relays have no pinned identity to preserve — they regenerate one per restart, which is fleet status quo, unchanged by this script.

**2. Empty token passed `update` verification (P1) — fixed, further than reported.** The check now requires a non-whitespace first character after `=` **and** rejects files containing an empty duplicate assignment. Both extensions matter and were empirically verified against docker 29.4: a CRLF line (`TOKEN=\r`) looks non-empty to grep but docker's env-file parser strips the trailing CR, leaving an empty token in the container; and on duplicate keys docker's last occurrence wins, so `TOKEN=x` followed by `TOKEN=` is also empty in the container. Unit tests cover all four cases (empty, non-empty, CRLF, empty-duplicate).

**3. Best-effort host-key verification (P1) — fixed, in every mode.** Not just the token-carrying ones: verification found that a tokenless `update` falling back to `accept-new` would *launder* an unverified key into `known_hosts`, where a later `convert` would treat it as verified. So **all modes** now refuse to contact a host whose key can be neither found in `known_hosts` nor pinned via the Lightsail API; `OPENRUNG_ALLOW_TOFU=1` is the explicit per-run opt-in. `UserKnownHostsFile` is pinned so the check and the connection can't diverge on a per-host ssh_config. Known limitation (fail-closed): the reverse lookup matches Lightsail instances by IPv4 `publicIpAddress` only, so IPv6/DNS targets need a manual pin or `ALLOW_TOFU` — noted in the error message and README.

**4. Rollback hint could destroy a healthy relay (P2) — fixed.** The recreate is now three separately-reported stages: pull failure says "the live container was not touched" (no hint); swap failure prints restore-not-destroy guidance; only a post-swap failure prints the `rm -f` + `rename` hint. Verification pushed this further: the swap stage now **refuses to discard an existing `-old` backup while the "live" container is not running** (the signature of a previously failed roll — proceeding would have deleted the last healthy relay), and `update` requires a container to exist before rolling (previously `live=''` on verify-timeout produced a hint that destroyed the only container and restored nothing).

## Adversarial verification of the fixes

Before pushing, a 7-agent pass (one refuter per finding + shell-semantics/security/operational hunters) tried to break the fixes. Everything high/medium it found is in this commit:

- `verify_foundation` now requires `Running` with `RestartCount` 0 — a container that registers and then crash-loops no longer verifies (docker `ps` lists `Restarting` containers and `docker logs` retains the old registration line, so the previous check passed) — and treats ssh errors during polling as transient rather than misdiagnosing a dead container and printing a rollback hint on a network blip.
- Remote **stderr** is stripped of control characters before reaching the terminal (`2> >(scrub_stream)` in `ssh_run`) — it was the one remote channel the allowlist/scrub regime didn't cover, usable for terminal-escape injection by a compromised host.
- The token source variables are `unset` after reading, so ssh/aws/ssh-keygen children never inherit the credential (consistent with `create`'s `env -u` rationale); the staged `.tmp` is removed by an EXIT trap on orderly failures; legacy-file cleanup distinguishes "absent" from "unreachable" with an explicit yes/no marker instead of failing open on an ssh error; `install -d` no longer re-modes a pre-existing parent directory (an `OPENRUNG_ENV_FILE=/etc/x.env` override would have chmod'd `/etc` to 700); and `create`'s wait loop surfaces the last ssh error on timeout instead of blaming cloud-init for an auth failure.

## Testing

44 offline checks, all passing: the previous end-to-end suite (guards, ordering, `bash -x` trace-leak, stubbed-`lightsail-up.sh` create flow) plus unit tests that source the script's functions and execute the **real remote command strings** locally against `ssh`/`sudo`/`docker`/`install` shims — covering preservation end-state (identity kept, managed keys rotated, liabilities dropped), the dual-file merge, all four token-precheck cases, all three recreate-stage failure modes including the backup-protection refusal, strict pinning in every mode, crash-loop verification failure, and a full `convert_host` run with legacy-file migration. `shellcheck` clean (2 intentional info-level SC2029). Still not exercised against prod, per deploy authority — the suggested first live step remains `update` against a single relay, which now also proves the state-check hardening.

openrung-admin:
Codex round 2 addressed in 0d977d0 — all three confirmed and fixed:

**[P1] TOKEN export attribute** — real; bash keeps the export attribute of an inherited variable across assignment, so an operator with a generic `TOKEN` exported would have leaked the bearer into every child's environment. Fixed with `unset TOKEN` before the assignment, so the variable is recreated unexported. Covered by a unit test that exports `TOKEN` into the script's environment, loads a bearer, and asserts a child process sees nothing.

**[P1] Backup guard too weak** — real; a failed rollout can be running yet never register, and the previous guard (Running-only) let a re-run discard `openrung-relay-old`, the last known-good relay. The swap now refuses to remove an existing backup unless the current container has itself **registered and stayed up** (`registered relay` in its logs, `Running` with `RestartCount` 0 — the same bar `verify_foundation` sets). On refusal the rollback command IS printed, since restoring the backup is the correct recovery in exactly that state. Unit tests cover dead-live, running-but-unregistered, and verified-healthy (proceeds) cases.

**[P2] Stranded `relay.env.tmp`** — real; the EXIT trap only exists in the install session. `stage_preserved_env` now removes the tmp itself when its validation fails, and the identity readback removes it (best-effort over a fresh ssh) before dying. The remaining window is a hard connection drop mid-append, already noted in the comments; the next staging truncates it.

50 offline checks now, all passing; shellcheck still clean (2 intentional info-level SC2029).

openrung-admin:
Implemented in `cbfbff9`.

What changed:

- Reduced automated entry states to one canonical steady state: a running `openrung-relay` with no legacy container or env, `-old` backup, `-new` candidate, staged `.tmp`, or unrecognized relay/volunteer-suffixed container.
- Added fleet-wide read-only preflight before token retrieval or image pull. Unsupported or ambiguous state now causes no remote relay mutation and prints inspection plus manual migration/recovery instructions.
- Reworked rollout as an explicit transaction: preserve the live container as stopped `openrung-relay-old`, run and verify `openrung-relay-new`, then promote it and remove the old container in one guarded commit step.
- Added `rollback`, limited to the two exact interrupted-roll states: stopped old only, or stopped old plus candidate. All other combinations remain manual-only.
- Tightened fail-closed behavior around Docker state snapshots, token export, SSH host-key persistence, and malformed provider key responses.
- Added `foundation-up_test.sh` and a required `relay-deploy-script` CI job. The suite evaluates all 2,048 state combinations for each mode and executes the real multiline Docker transition/rollback commands against a file-backed simulator, including injected failures.

Validation: 64 local assertions passed, including shell syntax and workflow YAML checks. The new deploy-script check and all existing Go, Docker, punchcore, and Pages checks are green.

PR #77, opened 2026-07-16
Add independent component versioning

openrung-admin:
## Summary

- give the standalone relay, relay hub, broker, and volunteer desktop independent semantic version sources
- report relay artifact identity as `relay/<version>` in direct and tunnel registrations while retaining the existing compatibility alias
- add `--version`, startup version/revision logs, linker injection, and OCI image metadata
- split server releases into `relay-v*`, `relayhub-v*`, and `broker-v*` tags; keep `main` and immutable `sha-*` image tags
- make the volunteer desktop backend and About screen consume one `desktop-volunteer/VERSION` source
- document the separation between application build identity and API/tunnel/punch protocol compatibility

## Compatibility

- no broker API version or tunnel/punch protocol version changes
- `volunteer_version` remains the deprecated v1 compatibility alias
- relay, relayhub, broker, desktop, and mobile release numbers no longer imply lockstep deployment

## Verification

- `go test ./...` and `go vet ./...` for the root module
- `go test ./...` and `go vet ./...` for punchcore
- volunteer Go tests and vet
- volunteer frontend: 21 tests and production build
- source and linker-injected version CLI checks for relay, relayhub, and broker
- Docker Buildx checks for all three Dockerfiles
- all three Go 1.25 Docker build stages compiled with version/revision arguments
- workflow YAML parsing and component tag/version simulations
- `git diff --check`

## Review focus

- component/version naming and release-tag boundaries
- whether `relay/<version>` is the desired long-term relay directory representation
- GitHub Actions behavior for main, workflow-dispatch, matching release tags, and rejected mismatched tags
- rolling-deployment compatibility between standalone relays, volunteer desktop relays, relay hubs, and brokers

## Local packaging note

A native macOS Wails package build reached the final native linker and failed on `_OBJC_CLASS_$_UTType`. The changed Go modules and volunteer frontend compile and test successfully; the failure appears in the local Wails/Xcode native-link step.


openrung-admin:
## Review

The design here is right: build identity separated from protocol compatibility, one `VERSION` per component, docs that say so explicitly. Most of this is ready. I verified the mechanics rather than reading only, and two things need fixing before the first real release.

### Verified working

- `--version` works for all three binaries; ldflags injection overrides the embedded default (`relay/9.9.9 revision=deadbeef`).
- `go test version.go version_test.go` — the file-list-mode CI command — works; `//go:embed` resolves correctly in that mode.
- `npm ci` / `test` / `build` pass after the `version` removal from `package.json`; `__APP_VERSION__` is fully replaced and `0.1.0` is inlined into the bundle.
- **The `relay_version` change from `"dev"` to `"relay/0.1.0"` is safe.** Nothing parses, compares, or feature-gates on that string anywhere — broker, clients, selector, or dashboard. The Postgres column is unbounded `text`, so no truncation. Decisively: `desktop-volunteer` already shipped `"desktop-volunteer/0.1.0"` in production, so the `component/X.Y.Z` shape has been flowing through the broker all along.
- The `paths:` filter alongside `tags:` does **not** block release-tag builds — I checked this specifically because it looked like a trap. GitHub's docs are explicit: "Path filters are not evaluated for pushes of tags." Release tags trigger correctly.

### 1. `sha-*` tags are called immutable, but the release build mutates them

The PR body says it will "keep `main` and immutable `sha-*` image tags." That doesn't hold. Building `deploy/broker/Dockerfile` twice from `b22af34`, exactly as the workflow would:

| Build | `--version` output | Image ID |
| --- | --- | --- |
| main build | `broker/0.1.0-dev+sha.b22af34` | `69e6341b87f3` |
| release-tag build | `broker/0.1.0` | `ec45c092a9c6` |

Both get pushed to the same `sha-b22af34` tag, because `type=sha` carries no `enable=` guard. So tagging a commit that `main` already built — the documented release flow, steps 2 then 3 — silently replaces `sha-<commit>` with different bits.

This matters because the fleet pins `sha-*` tags in practice (prod broker is on `sha-5c413ac`, relayhub on `sha-3cfa1c8`), so those pins stop being reproducible.

Fix: gate `type=sha` on `is_release != 'true'`. Release builds then publish `X.Y.Z` (plus the digest), dev builds publish `main` + `sha-*`, and each commit gets at most one `sha-*` push.

### 2. `volunteer-v*` tags collide with `desktop-release.yml`'s `v*` glob

GitHub's `*` matches any run of non-`/` characters, so `v*` matches `volunteer-v0.1.0`. Pushing that tag triggers `desktop-release.yml` as well, and its release job (`if: startsWith(github.ref, 'refs/tags/v')`) also passes. Two workflows would then publish to the same GitHub Release — one naming it "OpenRung Desktop volunteer-v0.1.0" with desktop-client artifacts, the other "OpenRung Volunteer volunteer-v0.1.0".

This is pre-existing, but it has never fired because no `volunteer-v*` tag exists yet (only `v0.1.0`–`v0.1.2` and `punchcore/v0.1.0`). This PR is what makes that release path real, documented, and about to be used — so it breaks on first use.

Fix: narrow `desktop-release.yml` to `tags: ["v[0-9]*"]`, which still matches `v0.1.0` and leaves the three existing tags valid, but no longer swallows `volunteer-v*`.

Related: `docs/versioning.md`'s table omits the desktop client — the one component still holding the unprefixed `v*` namespace. Worth documenting alongside the others so the boundary is visible.

### 3. VERSION validation is enforced in one component out of four

`desktop-volunteer/version.go` validates and `log.Fatal`s on a malformed VERSION. But `resolvedVersion()` in relay, broker, and relayhub silently degrades to `"dev"` and ships. CI catches this on `main`, so it's mostly a local/dev-build gap — but the asymmetry is surprising when all four read the same kind of file.

### 4. Duplication worth consolidating

`resolvedVersion` / `resolvedRevision` / `versionInfo` are copy-pasted verbatim across all three `cmd/*/main.go`, and the semver regex now appears in 8 files (`.go`, `.ts`, and 5 workflows). A small `internal/buildinfo` package plus one reusable workflow step would collapse both — and would fix item 3 for free.

### 5. Minor

- The relay logs its version *after* `resolveAutoMode()` does network probing, and `run()` can return early before reaching it — so the version is missing from exactly the failure logs where you'd want it. Log it right after `flag.Parse()`.
- `resolvedRevision()`'s `return "unknown"` fallback is unreachable: `revision` is already initialized to `"unknown"`.
- The new `volunteer-frontend.yml` runs 21 tests, none asserting the `__APP_VERSION__` injection it exists to protect. One assertion that `APP_VERSION` matches `/^\d+\.\d+\.\d+$/` would cover the wiring.
- A `broker-v*` tag on a commit *not* on `main` still triggers a full release build — `branches: [main]` doesn't constrain the tag path.
- `relay_version` has no length or charset guard in `validateRegisterRequest`, while sibling free-text fields get `NormalizeLabel` / `NormalizeNodeClass`. With anonymous registration enabled in prod, that's up to ~64KB of arbitrary bytes into the signed relay list. Pre-existing and not a regression, but now that the field has a documented shape, it's the natural moment for a cheap guard.

### Verdict

Items 1 and 2 are worth fixing before merge — both are cheap, and both bite during the first release this PR is designed to enable. Everything else is follow-up material. Core design and the Go/TS changes look solid.

openrung-admin:
Pushed c6acd2a fixing items 1 and 2 from the review above.

**1. `sha-*` mutation** — `type=sha` now carries `enable=${{ steps.version.outputs.is_release != 'true' }}` in all three image workflows, so each commit gets at most one `sha-*` push, from the `main` build. Simulated against the workflow's own version logic:

| Scenario | `build_version` | tags published |
| --- | --- | --- |
| push to `main` | `0.1.0-dev+sha.b22af34` | `main`, `sha-b22af34` |
| push tag `broker-v0.1.0` | `0.1.0` | `0.1.0` (no `sha-*` re-push) |
| push tag `broker-v9.9.9` | — | rejected, VERSION mismatch |

`docs/versioning.md` updated to match: `sha-*` is documented as a dev-build identity, and since nothing in GHCR enforces tag immutability anyway, it now says to pin a digest where an image must be guaranteed not to move.

**2. `v*` glob collision** — `desktop-release.yml` now matches `tags: ["v[0-9]*"]`. Verified against GitHub's filter-pattern semantics (`*` → any run of non-`/` chars):

```
v[0-9]*   MATCH  v0.1.0        (existing desktop tags stay valid)
          MATCH  v0.1.2
            -    volunteer-v0.1.0   <- no longer swallowed
            -    relay-v0.1.0 / broker-v0.1.0 / relayhub-v0.1.0
```

I also added the desktop client to the `docs/versioning.md` table, since it is the one component still holding the unprefixed `v*` namespace and the constraint (no new component tag may start with `v` + a digit) is worth recording where the tag scheme is defined.

Items 3–5 are untouched and left as follow-ups.

One thing surfaced while documenting the desktop row, out of scope here: `desktop/frontend/src/core/config.ts` pins `APP_VERSION = '0.1.0'` while the desktop client has released through `v0.1.2` — its About screen has been under-reporting for two releases. That is exactly the drift this PR's `VERSION`-file approach prevents, and a good argument for migrating the desktop client onto it too.

PR #79, opened 2026-07-16
desktop: release client version 0.1.3

openrung-admin:
## Summary
- make desktop/wails.json info.productVersion (0.1.3) the canonical desktop version
- inject it into the frontend, native macOS/Windows metadata, and packaged client telemetry
- validate exact vX.Y.Z release tags before the platform build matrix

## Verification
- npm test
- npm run build
- node --test desktop/scripts/versioned-wails-build.test.mjs
- go test ./... and go vet ./... (desktop module)
- go test/go vet internal client and telemetry packages
- macOS arm64 package build, code-sign verification, and bundle version check

After merge, tag the merged commit as v0.1.3 to publish the GitHub release.

### DIFF
--- a/.github/workflows/desktop-frontend.yml
+++ b/.github/workflows/desktop-frontend.yml
@@ -9,11 +9,15 @@ on:
     branches: [main]
     paths:
       - "desktop/frontend/**"
+      - "desktop/scripts/**"
+      - "desktop/wails.json"
       - "LICENSE"
       - ".github/workflows/desktop-frontend.yml"
   pull_request:
     paths:
       - "desktop/frontend/**"
+      - "desktop/scripts/**"
+      - "desktop/wails.json"
       - "LICENSE"
       - ".github/workflows/desktop-frontend.yml"
 
@@ -37,6 +41,10 @@ jobs:
 
       - name: Test
         run: npm test
+
+      - name: Test build version wiring
+        working-directory: desktop
+        run: node --test scripts/versioned-wails-build.test.mjs
 
       - name: Typecheck and build
         run: npm run build
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -22,7 +22,40 @@ env:
   SING_BOX_VERSION: ${{ github.event.inputs.singbox_version || '1.14.0-alpha.35' }}
 
 jobs:
+  version:
+    runs-on: ubuntu-latest
+    outputs:
+      product_version: ${{ steps.version.outputs.product_version }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - id: version
+        name: Validate desktop version and release tag
+        shell: bash
+        run: |
+          set -euo pipefail
+          if ! product_version="$(jq -er '.info.productVersion | select(type == "string")' desktop/wails.json)"; then
+            echo "desktop/wails.json info.productVersion must be a string" >&2
+            exit 1
+          fi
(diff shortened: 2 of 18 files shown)


openrung-admin:
## Review: solid, verified end-to-end. No blockers.

I verified the claims by building rather than reading, and pushed a revision (f5a0bdd) with the three follow-ups below.

### What I confirmed

- **Go injection works.** Built a probe against the desktop module: `client.AppVersion()` returns `dev` by default and `0.1.3` with the script's exact ldflags. `replace openrung => ../` does link `openrung/internal/client` into the desktop binary.
- **Native metadata works.** A real `wails build` produces `OpenRung.app` with `CFBundleShortVersionString = 0.1.3` and `CFBundleVersion = 0.1.3`.
- **Windows metadata holds** despite there being no `desktop/build/windows/`: Wails' `ReadFile` falls back to its *embedded* default `info.json`, which templates `{{.Info.ProductVersion}}`.
- **The "caller cannot override the version" guarantee is real.** Verified against the Go linker directly: with duplicate `-X` on one symbol, the last wins, so appending wails.json's value last genuinely defeats an override.
- Tests pass (18 vitest, 5 `node --test`, `npm run build`); the bundle carries `0.1.3` with no stale `0.1.0`; all three CI platform jobs route through the packaging scripts.

### This fixes more than the description claims

Wails defaults `productVersion` to `1.0.0` when absent ([project.go#L196](https://github.com/wailsapp/wails/blob/master/v2/internal/project/project.go)). I built `main` in a scratch worktree to confirm: its bundle is stamped **`1.0.0`**. So the shipped v0.1.0, v0.1.1 and v0.1.2 macOS bundles all claim to be version 1.0.0. Worth mentioning in the release notes — it's a user-visible fix, not just internal tidying.

### Follow-ups pushed in f5a0bdd

**1. The `-X` symbol path was an unenforced magic string.** `versioned-wails-build.mjs` names `openrung/internal/client.appVersion`, and nothing checked it still resolves. The Go linker **silently ignores `-X` for a symbol it cannot find** — the build succeeds and the variable keeps its default. So a rename of `appVersion`, a move of its package, or the desktop binary dropping the import would leave every release reporting `dev` to broker telemetry, with the About screen still showing the right number (the frontend reads wails.json independently) and every check green. That's the same silent drift this PR exists to fix, one layer down.

Added `scripts/versionprobe` plus a test that links it with the flags `versionedWailsBuildArgs()` really produces and reads the value back. It runs under `go-checks` rather than `desktop-frontend` because it needs a Go toolchain and must fire when `internal/client` changes. Mutation-tested: renaming `appVersion` leaves `go build` green and silent, and fails this guard.

**2. The version rule lived in three places.** `vite.config.ts` had its own copy of the semver regex and wails.json parsing, alongside the script's. Collapsed it onto the `readProductVersion()` the script already exports (−20 lines). The `jq`/`grep` copy in `desktop-release.yml` stays — different runtime, and its fail-fast-before-the-matrix job is worth the duplication.

**3. The entry-point check was symlink-fragile.** Node realpaths the ESM entry point but not `argv[1]`, so through a symlinked checkout the script skipped `main()` and **exited 0 without building** — after which the packaging scripts would ship whatever stale binary was in `build/bin`. Reproduced, then fixed with `realpathSync`.

Re-verified after the changes: both `node --test` files green, 18 vitest green, `npm run build` green, and a real `wails build` still stamps `0.1.3` into the bundle.

PR #102, opened 2026-07-28
Prefer direct TCP 443 in Volunteer Desktop

openrung-admin:
## Summary

- try ordered automatic direct candidates (TCP 443, then the configured alternate) and use RelayHub only after fail-safe reachability probing
- add explicit least-privilege Windows and Linux direct-connect setup, with a safe macOS packaging blocker instead of insecure elevation
- add setup/status UX, packaging guidance, documentation, and focused integration tests

## Migration and compatibility

Existing `listenPort=8443` installations retain that setting as the alternate automatic candidate while trying TCP 443 first. Direct-only mode continues to use exactly the configured port, and the generic relay engine/CLI keeps its single-port behavior unless ordered candidates are explicitly supplied.

## Validation

- `go test ./internal/relayruntime/... ./cmd/relay`
- `go test -race ./internal/relayruntime/engine`
- `cd desktop-volunteer && go test ./...`
- `cd desktop-volunteer && go test -race ./...`
- `cd desktop-volunteer/frontend && npm test` (38 tests)
- `cd desktop-volunteer/frontend && npm run build`
- relevant `go vet`, build-tag cross-compilation, macOS Wails build/codesign verification, script syntax, diff, and formatting checks

## Platform validation limitations

Windows UAC/firewall and Linux polkit/file-capability flows were cross-compiled and tested with injected fakes, but not exercised on native packaged hosts. macOS direct setup is intentionally reported unavailable until a Developer ID-signed, notarized SMAppService helper can be packaged; no setuid, password piping, or root GUI workaround was introduced. Full macOS release packaging could not run locally because its Xray input was unavailable. No production infrastructure was changed.


openrung-admin:
Addressed both findings in `ee2eb6d`.

- Linux now normalizes legacy `getcap` `+ep` output to the canonical `=ep` form for exact comparison and removal. Parsing uses the final capability field, so capability-looking tokens in a whitespace-containing executable path cannot be accepted; unexpected additional grants still fail closed. Tests exercise post-Enable verification, privileged removal, Manager removal/restart behavior, and the adversarial path-token case.
- Read-only setup inspections now have a 45-second deadline at both the service boundary and Manager pre/post inspection path. `Start` uses a non-blocking lock attempt and returns an actionable retry message while setup is busy, then follows its normal start path once inspection exits. The Linux privileged helper separately limits its fixed post-authorization `getcap`/`setcap` work to 30 seconds.
- I intentionally did not put an outer timeout around Windows Enable/Remove: the unprivileged PowerShell launcher cannot reliably terminate an already-elevated child, so returning early could release serialization while the firewall rule is still changing. This uses the reviewer-suggested “time out only Inspect” option safely.

Validation passed: desktop `go test ./...`, race tests for `volunteerservice` and `directsetup`, `go vet`, Linux/Windows cross-compilation, and Linux directsetup runtime tests in a network-disabled container.

openrung-admin:
Windows packaging follow-up: native workflow run 30351646336 exposed that Wails' `bindings` generator executes on GitHub's elevated Windows runner, so the production Administrator guard incorrectly rejected the non-GUI generator. Fixed in `b5dde47` by excluding the guard only from the `bindings` build tag; production Windows binaries still contain and enforce the Administrator refusal.

Validation: desktop `go test ./...`, Windows bindings and production cross-builds, guard string present only in the production binary, and a successful native Windows packaging run: https://github.com/openrung/openrung/actions/runs/30351973789. The resulting portable ZIP includes the x86-64 GUI executable, verified Xray binary, and license/setup documentation.

openrung-admin:
Volunteer Desktop version is now **0.1.1** (commit `86208f8`). The version source feeds both the relay-reported component version and the About screen. Verification passed: `go test ./...`, 38 frontend tests, frontend production build, and the native macOS/Linux/Windows release workflow: https://github.com/openrung/openrung/actions/runs/30353337803.

PR #107, opened 2026-07-29
Add a relay-fleet page to the admin telemetry dashboard

openrung-admin:
## What

A dedicated **Relays** page on the admin dashboard (`/admin/telemetry/relays`, linked from the overview via a new nav), answering: how many relays are there, which are foundation vs volunteer, where are they, how many clients does each carry, how popular is each, and what software they run.

- **Fleet cards** — relays online, foundation / volunteer split, relays seen in the window but no longer registered, clients currently connected through any relay (distinct, not a per-relay sum), live sessions, advertised session capacity, distinct countries.
- **Panels** — online relays by country, relay version distribution, fleet mix (direct/tunnel, punch-capable, WSS-fronted).
- **Per-relay table** — label coloured by broker-attested class with FND/VOL marker, online/offline status, exit city+country, public endpoint, transport+capabilities, version, connected clients, active sessions vs capacity, window clients/sessions, success/failed attempts with modal failure reason, speed-test averages, last heartbeat (online) or last telemetry event (offline), registration time. Online first, busiest first.

## How

- New `TelemetryQuerier.TelemetryRelayStats` implemented by **both** backends: the in-memory reader aggregation and a single-statement Postgres query reusing the shared `events`/`sessions` CTEs (each materializes once; a sentinel row carries the overall distinct connected-client count, which per-relay rows can't express).
- `GET /admin/api/telemetry/relays?window=` merges those stats with the live registry (`store.List`). Offline relays keep the node class retained with their telemetry; an online row's descriptor class always wins. The response exposes only admin-relevant descriptor fields — never lease tokens, identity keys, or `exit_host`.
- Windows, auth, no-store, and CSP posture identical to the existing pages.

## Testing

- Unit tests for the aggregation semantics (failover attribution, modal-reason tie-break, active counting, retained class, window bounds) and the registry/telemetry merge (totals, IPv6 endpoints, class precedence, offline rows).
- HTTP tests: auth gating (401 API / login redirect), invalid window, full JSON round-trip, page + CSP, cross-links.
- **Postgres↔in-memory parity**: `TestPostgresTelemetryQuerierMatchesInMemoryRelayStats` requires byte-identical JSON across 1h/24h/7d on the shared fixture, plus the empty case. Full suite green against Postgres 16.
- Verified live in a browser against a locally seeded broker (4 registered relays incl. one unlabeled, plus a telemetry-only offline relay).

Also fixes the `broker-dashboard` launch config, which couldn't start since relay-list signing became mandatory (adds a dev signing key + foundation token).


openrung-admin:
**Fixed in 6a4d7c8** — agreed on both halves of the finding, and the fix uses the trust anchor the pipeline already had.

**Attestation gate.** Ingestion stamps `relay_node_class` onto a record only when the named relay held a live registration at receipt (`attestTelemetryRelayNodeClasses`); clients post `TelemetryEvent`s and never control the record field, so the stamp cannot be forged. `TelemetryRelayStats` now returns rows only for IDs with at least one stamped record in the window, on both backends (`WHERE e.node_class <> ''` in SQL, the mirrored check in Go), and the fleet-wide connected-clients sentinel honours the same gate — fabricated heartbeats no longer inflate it. Events received during a brief registration gap still count toward an attested relay's aggregates (only rows are filtered, not events). The parity fixture now contains an unattested ghost relay (with an active session and a speed test) and a registration-gap record, and the byte-identical parity test pins both backends to the same filtering.

**Bounded response.** As a second layer, offline rows cap at the busiest 200 per response (`maxOfflineRelayRows`); totals still count every attested offline relay and the page shows "Showing the busiest N of M offline relays" when the cap hides rows. Online rows stay uncapped — they mirror the live registry, which expires unrenewed leases within minutes.

**Same-class hole in the overview.** The speed-tests panel also emitted one row per arbitrary `relay_id` with no cap; it now ranks top-ten like every other relay-keyed panel, with relay-ID tiebreaks (also added to the top-relays sort) so truncation is deterministic across backends.

**Residual, deliberately out of scope.** With anonymous registration enabled, an attacker can still *register* relays (separately rate-limited, logged, endpoint-keyed) and thereby mint attested IDs — that is the same trust boundary as the public relay directory itself and predates this PR. Likewise, hostile cardinality still costs the SQL `GROUP BY` work even though no rows are returned for it; that ingest-side concern applies to several existing dashboard aggregations and is best handled together.

Verified end-to-end against a local broker: a 50-ID forged flood through the telemetry API appears nowhere on the page, while a relay that registered, carried real sessions, and let its lease lapse still shows as an offline row with its retained class.

openrung-admin:
**Fixed in 87dec98** — good catch; I had accepted the lease-gap zeros as a corner case, but you're right that the current registration is itself broker attestation and closing the gap costs nothing security-wise.

The handler now reads the registry first and feeds its relay IDs into `TelemetryRelayStats` (new `registeredIDs` parameter on the querier). Both backends keep a row when the ID is **currently registered or historically stamped** (`WHERE e.node_class <> '' OR e.relay_id = ANY($5)` in SQL, the mirrored `trusted()` check in Go), and the connected-clients sentinel honours the same pair — so an online relay's gap-received failure reports now surface instead of rendering zeros. Because the same `store.List` result both vouches for the rows and drives the merge, a relay can't be trusted by one side and missed by the other.

Telemetry-only offline IDs still require the retained stamp, so the P1 property is unchanged: fabricated relay IDs from the anonymous telemetry API appear nowhere, and the registered set comes from the broker's own store, never from telemetry. Covered by a registered-but-unstamped relay in the parity fixture (byte-identical across backends and windows), a gap-relay case in the HTTP test registered end-to-end, and the unit fixture's registered/stamped/fabricated trio.

PR #108, opened 2026-07-29
volunteer-up.sh: generated relay names, one-line custom naming, thank-you banner

openrung-admin:
## What

Volunteer-experience improvements to `volunteer-up.sh` — strictly non-interactive throughout:

1. **Every relay gets a name.** The first run generates an adjective-noun name from the same word lists as `lightsail-up.sh`/`hetzner-up.sh`, so volunteer relays sit alongside fleet names in the dashboard instead of showing as raw relay-ID hex (previously the one-liner set no label at all). Names come from `/dev/urandom` via POSIX `cksum` (no `$RANDOM` in dash/ash). A custom name is chosen in the one-line command itself:
   ```sh
   curl -fsSL https://raw.githubusercontent.com/openrung/openrung/main/deploy/relay/volunteer-up.sh | sudo env OPENRUNG_LABEL=my-relay sh
   ```
   The script never prompts, so cloud-init and CI can run it as-is. The label is now always written to `relay.env` on first run; the env file stays authoritative on re-runs.
2. **Banner + thank-you closer.** Successful runs end with the lowercase `openrung` figlet banner (standard font, 47 columns — fits any 80-column terminal) and, as the literal last line, bold when stdout is a terminal (plain when piped):
   > thank you for running 'merry-falcon', together we will make the internet open again

Also: `install -d` derives the directory from `ENV_FILE` (as `foundation-up.sh` does) instead of hardcoding `/etc/openrung`, which lets the test suite's path rewrite reach the first-run branch.

## Tests

- Suite grows **41 → 50 assertions**: new first-run coverage (generated label written and matching the adjective-noun shape, explicit `OPENRUNG_LABEL` honored verbatim, candidate promoted with no stop/rm of a nonexistent prior container) plus banner-present and exact-final-line assertions on both the first-run and update paths. Green under dash, bash, and busybox ash (alpine), same CI matrix.
- Verified on a real pty (`script(1)`, stdin carrying the script exactly like `curl | sh`, typed input waiting): no prompt ever appears, the generated name is used, and the bold escape renders only on the terminal.
- `shellcheck -s sh` clean on both files.

`deploy/relay/README.md`'s one-liner section documents the naming behavior.

Per our process: ready for Codex review, then user merge.

### DIFF
--- a/deploy/relay/README.md
+++ b/deploy/relay/README.md
@@ -241,7 +241,13 @@ root-owned mode-`0600` `/etc/openrung/relay.env`, registration against the
 broker's direct TLS origin — except that no `OPENRUNG_FOUNDATION_TOKEN` is
 installed, so the broker attests the relay volunteer-class. It auto-detects
 the public IPv4 address, mints a stable `OPENRUNG_IDENTITY_SEED` once, and
-verifies the relay registers before declaring success. Re-running the same
+verifies the relay registers before declaring success. On an interactive
+first run it asks for a relay name (public in the relay directory), defaulting
+to a generated adjective-noun name in the fleet style; the prompt reads
+`/dev/tty` and is skipped automatically when there is no terminal, so
+cloud-init and CI never hang — set `OPENRUNG_LABEL` to name the relay
+explicitly, or `OPENRUNG_NONINTERACTIVE=1` to force the generated default.
+Re-running the same
 line is the update path: the image is pulled, the serving container is retained
 as a stopped backup, and a separately named candidate must register and remain
 running without restarts before it is promoted. A failed start, crash loop, or
--- a/deploy/relay/README.md
+++ b/deploy/relay/README.md
@@ -241,13 +241,11 @@ root-owned mode-`0600` `/etc/openrung/relay.env`, registration against the
 broker's direct TLS origin — except that no `OPENRUNG_FOUNDATION_TOKEN` is
 installed, so the broker attests the relay volunteer-class. It auto-detects
 the public IPv4 address, mints a stable `OPENRUNG_IDENTITY_SEED` once, and
-verifies the relay registers before declaring success. On an interactive
-first run it asks for a relay name (public in the relay directory), defaulting
-to a generated adjective-noun name in the fleet style; the prompt reads
-`/dev/tty` and is skipped automatically when there is no terminal, so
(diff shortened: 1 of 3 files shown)


openrung-admin:
Pushed a follow-up: successful runs now close with the lowercase openrung figlet banner and the reworded bold final line — `thank you for running '<name>', together we will make the internet open again` (bold only when stdout is a tty; plain when piped). Suite is now 50 assertions, still green on dash/bash/busybox ash; verified rendering on a real pty.

openrung-admin:
Reworked per discussion: the /dev/tty prompt and OPENRUNG_NONINTERACTIVE are gone — the script is strictly non-interactive again. Custom names are passed in the one-line command (`sudo env OPENRUNG_LABEL=my-relay sh`); otherwise a fleet-style adjective-noun name is generated. Banner + bold final thank-you unchanged. PR title/body updated; 50 assertions still green on dash/bash/busybox ash.

openrung-admin:
Both review findings confirmed against the source and fixed in the latest commit:

**63-character limit** — confirmed: `relay.NormalizeLabel` (internal/relay/types.go, `MaxLabelLength = 63`) errors on longer labels and `cmd/relay` exits at startup, so a persisted over-long label crash-loops every later run. `validate_label()` now mirrors the binary's rules (slug charset + ≤63) and refuses before anything is persisted, with the limit named in the error. Note the prompt half of the finding is already moot — the interactive prompt was removed in an earlier commit; naming is env-var-only.

**Unnamed legacy relays** — confirmed: `cmd/relay/main.go` calls `relayruntime.GenerateLabel()` when the label is empty, i.e. per process start, so pre-naming env files churned their dashboard name on every restart. The reuse path now backfills a pinned label when the env file has no usable one (last-occurrence-wins, matching docker `--env-file`): the `OPENRUNG_LABEL` override when given, else a generated name — appended atomically after the Foundation-ownership refusal, so Foundation files are never touched, and identity lines are never rewritten. A named env file remains authoritative; later overrides are ignored with an explicit note.

Suite: 50 → 56 assertions (backfill on the update path, 63-char boundary pinned verbatim, named-env override ignored without duplicate lines, 64-char refusal before any Docker mutation), green under dash, bash, and busybox ash; shellcheck clean.

PR #109, opened 2026-07-30
Gate telemetry ingest and cap anonymous relay registrations

openrung-admin:
Follow-up to #107, which bounded the admin relay dashboard's *responses* against fabricated relay IDs. This PR addresses the two residual issues on the *write* paths: hostile high-cardinality telemetry still reaching storage (and every dashboard GROUP BY), and anonymous registration minting attested relay identities without bound.

## 1. Telemetry ingest gating

**Unknown-relay discard.** Events carrying a `relay_id` that is neither actively registered (i.e. attestation stamped a `relay_node_class`) nor present in a new 7-day **recently-registered ledger** are discarded before they reach the sink, the relay-metrics ranking, or `RecordRelayTelemetry`. Such an ID cannot have appeared in any signed relay list a client saw, so the event is fabricated by construction.

- The "relay just died" trade-off flagged in the task is resolved by the ledger, not by requiring an active lease: any relay registered or heartbeated within the telemetry retention window (7 days — the widest window any dashboard reads) stays reportable after death. The ledger refreshes on **every heartbeat**, so a relay that registered weeks ago and only ever heartbeated keeps coverage.
- The ledger is process-local, bounded (16k entries, ~1 MB worst case), and seeded from the store's active descriptors at startup, so a broker restart on the Postgres store forgets only relays that died *before* the restart and never returned.
- Malformed IDs are rejected even cheaper: every ID the broker has ever minted (legacy random and identity-derived) is `relay_` + 32 lowercase hex, now codified as `relay.WellFormedRelayID` next to `DeriveRelayID`. The same check now guards the heartbeat path ID and WSS ticket `relay_id` lookups.
- Dropping is silent per-event (the `accepted` count reflects stored events, with a per-batch `slog.Warn`), **not** a batch 400 — a fielded client with one odd event must not get its whole retry outbox stuck.

**Poison-batch fix (found while implementing).** A single event containing NUL or invalid UTF-8 made the async Postgres insert fail; the failed batch is re-queued at the front of the pending buffer, so one hostile event could wedge telemetry flushing until 16 MB of backlog forced it out (plus duplicate rows from retried sibling chunks). Now:
- `validateTelemetryEvent` rejects (400) events whose string fields/attribute keys/values are not valid UTF-8 or contain NUL. Other control characters stay allowed in attributes (multi-line `failure_detail` is plausible).
- `recordClientSeen` was a side door: it builds events from raw HTTP headers, which may legally carry non-UTF-8 bytes (obs-text) that the JSON decoder would have replaced. Unstorable client/session IDs skip the record; unstorable version headers degrade to empty.

**What this deliberately does not fix:** an attacker minting junk `city`/`ISP` strings on relay-less session events. Those group by session first, volume is bounded by the existing rate limits, and the values are inherently client-asserted; distinct-value caps per `client_id` were considered and rejected (attacker mints client IDs as freely as cities, so the cap adds complexity without a bound). Server-attested geo from the source IP would be the real fix and is left as future work.

## 2. Registration hardening

**Long-window new-identity cap.** Beyond the short-window rate limiter (20/s — sized for hub bursts, i.e. ~1.7M/day from one IP), the broker now caps **successful registrations of new relay identities** at 64 per source IP (IPv6: per /64) per rolling 24 h:

- A registration re-proving a **known** identity is a reconnect, not a new relay — tunnel relays rejoin through the hub constantly and stable-identity relays re-register on every restart; both bypass the cap via the same ledger.
- Budget is consumed **only after the store accepts** the registration, so a relay crash-looping on a bad proof or WSS misconfig cannot burn its own budget while the operator debugs (checked with a test).
- Over-cap answers 429 + `Retry-After: 3600`, which `registerWithRetry` in the engine already honors — a throttled legitimate volunteer recovers unattended.
- Escape hatches: `OPENRUNG_MAX_NEW_RELAY_IDS_PER_IP_PER_DAY` (positive integer or `off`), `OPENRUNG_REGISTRATION_CAP_EXEMPT_CIDRS` for infrastructure that legitimately fronts many first-time registrants, and an unconditional bypass for the foundation token.
- The default (64/day) is sized for the worst legitimate shape: a flapping **legacy** (identity-less) relay that mints a fresh random ID on every successful re-register.

**Field validation.** `public_host`/`server_name`/`exit_host` must be syntactic hostnames or IP literals; the remaining free-form fields get length caps and a no-control-characters rule. These values are served verbatim to every client inside the signed relay list, so free-form strings were both a response-inflation and a display-injection surface.

**Evaluated and deferred** (documented here rather than half-built):
- *Reachability probes before listing*: a TCP probe can be pointed at third parties (register `1.1.1.1:443` → broker + clients hammer someone else), and a probe deep enough to verify a Reality endpoint means implementing a Reality client in the broker. Also risky for onboarding if a relay registers before its xray listener is up. Worth a separate design (async verify + grace period + per-endpoint probe dedup).
- *Proof-of-work*: only rate-limits identity minting per CPU, which the per-IP cap already does more cheaply; a difficulty that deters a state-level botnet also taxes volunteers. Revisit if capped-per-IP minting across many IPs becomes an observed problem.

## Prod deploy notes

- Set `OPENRUNG_REGISTRATION_CAP_EXEMPT_CIDRS=43.201.124.63/32,43.201.172.102/32` (relay hub EIPs) in `broker.env` so a viral wave of first-time tunnel volunteers is never throttled. Steady-state rejoins are uncapped even without this.
- No schema changes, no client changes, no migration. The ledger warms from the store + heartbeats within one lease TTL of startup.

## Interplay with #107

Touches `internal/broker/server.go` (handler wiring) but not the dashboard files or query CTEs, so conflicts with #107 should be trivial. Note the CTE-cost concern from the task is addressed at the root: once fabricated rows stop being stored, the GROUP BYs' key space converges to real fleet cardinality as hostile rows age out (≤7 days).

## Testing

- New: ledger unit tests, cap unit + full-handler tests (429/Retry-After, known-identity and foundation bypass, exempt CIDR, failed-attempts-don't-consume), ingest gating unit + end-to-end (fabricated dropped, attested kept, accepted count), NUL/UTF-8 rejection incl. the header side door, `WellFormedRelayID` vectors, register field-hardening table.
- `go vet ./...` clean, full `go test ./...` green, `gofmt` clean.


openrung-admin:
Review findings addressed in 49b8005 — all five confirmed and fixed:

- **[P1] Concurrent cap bypass** — replaced check-then-record with an atomic `reserve`/`release`: the unit is claimed under the lock before the store call and handed back on every failure path (keeping the "rejected requests don't consume budget" property via a deferred release). `TestRegisterCapHoldsUnderConcurrency` fires 80 simultaneous registrations at limit 4 and asserts exactly 4 are created; passes under `-race`.
- **[P2] Foundation consumed anonymous budget** — the reservation is now only taken when `newIdentity && maxClass != foundation`, so foundation registrations neither hit the cap nor spend the address's budget. Pinned by `TestRegisterCapFoundationDoesNotConsumeAnonymousBudget` (foundation first, then anonymous still gets 201).
- **[P2] Fixed window advertised as rolling** — the cap now keeps per-source reservation stamps and prunes by `now - 24h`: an exact sliding window, so a trailing 24-hour interval can never see more than the limit. Memory stays bounded at <= limit stamps (8 B each) per tracked source; the sliding-window property is asserted in `TestNewRelayCap` (the unit freed at hour 25 is exactly the one reserved at hour 0, not both).
- **[P2] Just-expired relays invisible on the relays page** — the ledger now retains the node class attested at the relay's last registration/heartbeat, and `gateUnknownRelayTelemetry` stamps it onto ledger-approved records exactly like live attestation would. A relay that just died keeps its class-filtered dashboard attribution for the full 7-day window (re-registration under a new class updates the retained class, covering volunteer->foundation conversions).
- **[P3] C1 controls accepted** — `registerTextValid` now uses `unicode.IsControl` (the whole Cc category, U+0080 through U+009F included). Test case added with U+009B via the "backslash-u009b" escape.

Docs updated to match (retained-class attribution sentence; the "rolling 24 h" wording is now literally true). Full suite + vet green.

openrung-admin:
[P2] same-identity duplicate races — fixed in b23d16f, using the "atomically remember if new after success and release duplicate reservations" option:

- `relayIDLedger.rememberIfNew` commits the identity and reports, under the ledger lock, whether this sighting is the one that (re)established it. The register handler keeps its reserved unit only on a true verdict; commit-race losers hand their units back through the existing deferred release. One identity can now consume at most one unit no matter how many concurrent duplicates race, while distinct identities (legacy random IDs included) still charge normally.
- Fail-closed corner: when the ledger is at capacity and cannot record the newcomer at all, `rememberIfNew` still reports true, so the reserved unit stays consumed — the cap must not open up under exactly the cardinality pressure that fills the ledger.
- Known trade-off vs. serializing by derived ID: while duplicates are in flight they transiently hold multiple units, so a burst of more than `limit` simultaneous copies of one identity can see some of its own duplicates 429 until the first completes (they are registering a relay that is already being registered; the engine's Retry-After backoff recovers). The settled state is exact, which is what the finding measured.
- Tests: `TestRegisterCapChargesConcurrentSameIdentityOnce` reproduces the reported scenario deterministically — `limit` concurrent registrations of one unseen key must all return 201, then `limit-1` further distinct identities must still fit and the one after that must 429 (i.e. the race consumed exactly one unit). Passes with `-race -count=5`; run against the pre-fix handler it fails as the finding described (the duplicate race drained the budget and the second distinct identity already received 429). `TestRelayLedgerRememberIfNew` covers the verdict semantics including refresh, expiry re-establishment, and the at-capacity fail-closed path.

Full suite + vet green.

PR #120, opened 2026-08-05
Provision and advertise the Azure Front Door broker front

openrung-admin:
Superseded by #121.

The work was rebased onto current `main` and re-opened as a single clean commit. This PR is closed and its branch deleted.

See #121 for the Azure Front Door broker front.


openrung-admin:
## Endpoint renamed before merge — `cdn-edge-cxdnhsg2aadmaubj.z02.azurefd.net`

The endpoint was originally provisioned as `openrung-broker-<hash>.z02.azurefd.net`, which undercuts the point of the front.

Suppressing SNI keeps the endpoint name out of the ClientHello, but the client still resolves it over **ordinary cleartext DNS** — `brokerapi` installs no custom resolver and there is no DoH in the path. The hostname is therefore the one part of this front a passive on-path observer still sees, which `docs/architecture.md` already stated: *"DNS resolution of the distribution name still exposes it."*

A project-identifying prefix makes that query a keyword match. It lets the front be blocklisted by pattern with no prior knowledge of the project, and it **marks the user** as running this software to anyone logging their DNS — a personal-safety property in the places this matters, not just a reachability one. CloudFront's `d2r7mdpyevvs1m.cloudfront.net` reveals nothing, and this now matches that.

Front Door endpoint names are immutable, so this is a new endpoint and route. It was verified before the old one was deleted, and the gate passes:

```
PASS  transport TLS policy for this host
PASS  handshake completes and satisfies that policy
PASS  signed relay list fetches and verifies over the shipping path
        4119 bytes, key 627405615601c589, server_time 0s old
        carried over TLS with ClientHello server name ""
PASS  SNI-bearing control serves the same signed list
PASS  edge routes on the Host header, not the address
```

Nothing had shipped, so no client ever saw the old name.

### Also added: a propagation trap

A new endpoint spends several minutes returning an intermittent mix of 404 and 200 **from the same edge address**, because servers behind that anycast IP pick up the config at different times. Running the gate in that window fails only the two relay fetches while the handshake and Host-routing checks pass — which reads as a routing misconfiguration and isn't. The runbook now says to poll for steady 200s first.

### Not addressed here

The same reasoning applies more strongly to `broker.openrung.org`, the *primary* front: it is a subdomain of the project's own domain, and ECH hides the SNI but not the A-record lookup. A DNS-observing censor learns more from front #1 than from anything changed here. That is a larger, separate decision — the Cloudflare front is arguably the deliberately well-known one — but it is where the exposure is concentrated.

openrung-admin:
Superseded by #121 — rebased onto current main as a single clean commit.

PR #133, opened 2026-08-16
TUI: desktop connection parity (ADR-001 B2)

openrung-admin:
ADR-001 Track B2: bring `cmd/client` to functional parity with the desktop GUI's connection behavior — everything via `internal/connectcore` configuration, nothing reimplemented in the view.

## What's here

**1. Multi-front broker discovery.** `-broker` now defaults to empty, which means the engine's `brokerapi.BrokerCandidates`/`FirstReachable` staggered race over the built-in HTTPS fronts (Cloudflare → CloudFront → Azure-as-last-resort), exactly like desktop. A set `-broker` remains an override: per the engine's existing `OverrideFirst` policy it is attempted strictly first and **alone** — no default front is contacted while it is pending, and on any success it wins. `check`/`config` follow the same rule (explicit URL = single deterministic endpoint, unchanged code path; empty = the front race).

**2. Engine events surfaced in the UI.** New typed `Notice` events beside the existing `EventSink` (an optional `NoticeSink` interface — the desktop sink doesn't implement it and is untouched):
- failover re-ladder: `failover_started` (which relay died + why) and `failover_completed` (to which relay, via which front if WSS),
- WSS fallback engagement per front, carrying the direct-path failure, plus the bounded ticket Retry-After wait,
- punch attempt outcome (NAT class on success, reason + hub fallback on failure),
- every mid-session health probe sweep (consecutive failures out of threshold, incl. the "network looks down, not failing over" case).

The TUI renders these as **Health** and **Activity** rows on the Status view; log lines are unchanged, so no host loses information.

**3. Shell proxy helper.** Settings gains a "Shell proxy" action that generates and displays the two copyable POSIX commands from `internal/proxyconfig` (`. '<helper>' && openrung_proxy_on` / `openrung_proxy_off`), writing the same sourceable helper desktop's `GetProxyInfo` writes. Enabled only while connected, mirroring the desktop Settings gating; unavailable on Windows, like desktop.

**4. Recents.** The engine already persists recents through `internal/clientstate` (same on-disk state desktop reads); the TUI now seeds them at startup from `engine.State()` and renders them newest-first as a row in the Relays view.

**5. Telemetry platform label.** New `Engine.Platform` (empty = `PlatformDesktop`, preserving desktop wire behavior byte-for-byte) threads through the shared `clienttelemetry` manager, `internal/discovery`, and WSS ticket requests. The CLI sets `connectcore.PlatformCLI` (`"cli"`).

### ⚠️ New platform label in dashboards
Every telemetry event from the terminal client now carries **`attrs["platform"] = "cli"`** — this will appear in dashboards as a new attribute value alongside desktop/mobile sessions. Verified the broker tolerates it:
- `validateTelemetryEvent` treats attributes as free-form (≤32 entries, 64/256-byte caps, UTF-8) — confirmed by running the real validator against an event carrying `platform:"cli"`;
- dashboards read specific keys (`operating_system`, `app_version`, …) and ignore unknown ones;
- `brokerapi` sends **no** platform header for an unknown `Platform` value, so header-side ingest (`client_seen` reads only `app_version`/`android_api`) is unaffected.

## Resolved divergences

- **`check`/`config` default endpoint changed.** They previously defaulted to `http://localhost:8080`; they now race the production fronts when `-broker` is empty. Scripts that relied on the localhost default must pass `-broker http://localhost:8080` explicitly. (Keeping localhost as a *default* would have made the bare TUI unusable for real users, which defeats item 1.)
- **A failing `-broker` override still falls back to the built-in fronts.** That is the engine's (and desktop's) `BrokerCandidates` semantics: the override is exclusive while pending, but its failure re-enters the phased default race. The prompt's "narrows candidates to one" holds for a healthy override; making a dead override terminal would have required diverging from the shared engine policy.
- **CLI requests no longer claim the desktop platform header.** B1 shipped the TUI sending `X-OpenRung-Desktop` on discovery/ticket/telemetry requests (hardcoded in shared code). With `Platform="cli"` unknown to `brokerapi`, those CLI requests now send no platform header at all. Desktop requests are unchanged.
- **The platform attribute is stamped only for non-desktop platforms**, so desktop telemetry stays byte-identical. If we later want `platform:"desktop"` too, that's a one-line follow-up in `Engine.newManager`.
- `cmd/client/VERSION` bumped 0.2.0 → 0.3.0.

## Acceptance mapping

- *Forced direct failure engages WSS fallback visibly*: `TestWSSFallbackEmitsTypedNoticePerFront` (engine emits the notice with relay + front + direct failure; existing WSS ladder tests pin the fallback itself) + `TestEngineNoticesDriveActivityAndHealthRows` (the Status view renders it, and `ConnectionInfo` already shows `Transport: WSS front <id>` once promoted).
- *Mid-session relay kill triggers visible failover*: `TestMidSessionFailoverEmitsTypedNotices` (started + completed notices with reason and winner) + the same UI render test.
- *Telemetry arrives at a local broker with the new label*: `TestEngineTelemetryCarriesCLIPlatformLabel` runs the engine against a loopback broker and asserts `platform="cli"` on `connection_succeeded` and `connection_ended` (and its desktop-default subtest asserts absence).

A live smoke against a WSS-capable fleet relay (unreachable-IP pin → watch the Activity row engage the front) is worth running before release; it needs fleet access this environment doesn't have.

## Testing

- `gofmt` clean; `go build ./... && go test ./...` green in the root module, `desktop/`, and `desktop-volunteer/` (desktop modules built with the usual local `frontend/dist` placeholder).
- New tests: engine notice emission (failover, WSS fallback, ticket retry, punch, health loop), engine→loopback-broker platform label (both platforms), `clienttelemetry.SetPlatformLabel` (stamping, caller-wins, heartbeat, default-absent), TUI notice/health rendering + clearing on disconnect, recents row, shell-helper gating/success/error.


openrung-admin:
Addressed all three findings in 7c55e02:

1. **Shell commands after disconnect/failure** — rendering is now gated on connection state rather than clearing: the enable command (which points at the dead local endpoint) is replaced by "available while connected", while the restore command stays visible — a shell still carrying our proxy variables needs `openrung_proxy_off` precisely after disconnect/failure/crash, and desktop keeps its restore copy-button enabled the same way. Pinned by the extended `TestShellHelperActionRequiresConnection`.

2. **Activity leaking into the next session** — the activity notice is cleared on `preparing`, which only a fresh `ConnectTarget` passes through (failover recovery re-promotes via `connecting`), so it survives the terminal screen and mid-session recoveries but never a genuinely new connection. Pinned in `TestEngineNoticesDriveActivityAndHealthRows`.

3. **`4/3 probes failed`** — `healthLoop` now caps the notified count at `HealthFailureThreshold` (the internal counter still grows, and the "network looks down" reason still distinguishes the outage hold from a triggering failure). New `TestHealthLoopCapsNotifiedFailuresAtThreshold` drives a prolonged-outage sequence and asserts no notice ever exceeds the threshold.

gofmt clean, full root-module suite green.

PR #134, opened 2026-08-16
Retain full-device TUN mode in the TUI (ADR-001 B3)

openrung-admin:
Track B3 of [docs/adr/001](docs/adr/001-connectcore-and-tui.md): full-device TUN capture returns to the terminal client, and the docs are rewritten for the client that actually exists now.

## Capture mode

`connectcore.Mode` is a new engine-level setting — `ModeProxy` (the zero value, so every existing host keeps its behavior byte for byte) or `ModeTUN`. It is deliberately not `client.InboundMode`, whose zero value is `ModeTUN`: an `Engine` literal must never silently claim to need root.

`SetMode` refuses while a connection is live. The inbound shape, the OS-proxy policy, the probes, and the elevation requirement all follow from the mode, so a session keeps the mode it started with.

TUN mode uses `BuildSingBoxConfig`'s existing TUN shape unchanged: `auto_route`, `strict_route`, `dns_mode: hijack`, and `route_exclude_address` for the relay IP plus a punched peer. `-mtu` now reaches it (it was previously warned about as ignored).

## Elevation

The gate is A1's `Elevation` hook, invoked from a new `Engine.prepare` that both `Prepare()` and the connect flow run. It fires while the state machine is still `PREPARING` — before the telemetry session opens and before the broker is contacted — because a privilege refusal is a local precondition, not a connection attempt.

The CLI's implementation verifies rather than acquires: re-executing under `sudo` would detach the tunnel from the terminal that owns the TUI, so it checks euid (unix) or the process token (Windows) and returns the exact rerun command. A nil hook refuses too, rather than letting sing-box fail opaquely after the ladder has dialed relays.

```
$ go run ./cmd/client connect --tun -headless
status: preparing
[12:48:29] connect failed: TUN mode needs root privileges to create the tunnel device: rerun as `sudo client connect --tun`, or drop --tun to use proxy mode (no privileges needed)
status: failed
```

## What TUN mode changes inside the engine

A TUN inbound binds no port, so proxy-mode machinery is bypassed or replaced:

- **Readiness** — proxy mode dials the loopback mixed inbound. TUN mode waits for the config's tunnel address to appear on a local interface (`tunInterfaceReady`), which sing-box assigns before it installs routes.
- **Internet and health probes** — the same sweep/retry policy over a client with no proxy set (explicitly `nil`, never `ProxyFromEnvironment`, so an inherited variable cannot send the probe somewhere other than the tunnel being verified).
- **Stable proxy port** — not resolved, not reserved, not persisted. `ConnectionInfo.ProxyPort` is 0 and the TUI shows a TUN capture row instead of an endpoint.
- **OS proxy** — `applyProxy` is a no-op; pointing the system proxy at a port nothing listens on would blackhole the traffic the tunnel carries, and would leave a snapshot to restore for a change never made.
- **Kill grace** — `TUNKillGrace` (5s) replaces `LadderKillGrace` (500ms). A hard kill would leave the routing table pointing at a gone interface.

## Resolved divergences

Two engine behaviors could not hold in TUN mode. Both are consequences of a full-device tunnel owning the default route, both are documented in `docs/desktop-client.md`, and both have tests.

1. **WSS/CDN fallback is skipped in TUN mode.** The WSS bridge dials its front from the client process; the TUN captures that connection, sing-box routes it to the loopback bridge, and the bridge dials the front again — a loop. Only the relay's own address can be excluded from the routes, not a CDN's (whose addresses rotate and are not in the descriptor). A relay whose direct path fails is left failed, with a log line saying why, rather than looped.

2. **TUN mode fails over on the health threshold alone.** Proxy mode gates a failover on "is the local network alive at all", dialing broker fronts outside the tunnel, so a Wi-Fi blip or a laptop sleep does not churn relays. In TUN mode that reference point is reached *through* the tunnel under suspicion, so the gate could only ever answer "network down" — which would disable failover permanently. Skipping it means a genuine local outage ends the session instead of riding it out; the recovery pass tears the TUN down first, which is also what restores the normal network in that case. Fail-open, consistent with the ladder's existing availability-over-leak stance.

Neither affects proxy mode, the desktop app, or mobile.

## Manual test

**Not executed in this environment:** `sudo` here requires a password I cannot supply, so the full-device leg of the acceptance criterion was **not run**. What was verified, and the procedure for the sudo leg, follows.

Verified directly:

- **Non-sudo TUN is refused with the guidance**, via both `--tun` and `-tun`, before any broker contact (transcript above; a test asserts the broker fetch and sing-box start never happen).
- **The generated TUN config is accepted by the real sing-box.** The engine's `candidateConfigInput` output for `ModeTUN` (with `-mtu 1400` and a punch peer) was dumped and run through `sing-box check` against sing-box 1.14.0-beta.14 — `SING-BOX CHECK OK`, with the inbound carrying `auto_route`, `strict_route`, `dns_mode: hijack`, `mtu: 1400`, and both `route_exclude_address` entries.
- **Proxy mode is untouched**: a headless connect still runs its normal ladder with no elevation call, and leaves no system proxy enabled on exit.
- `go build`, `go vet`, `go test` clean in the root, `desktop/`, and `desktop-volunteer/` modules; `GOOS=windows` and `GOOS=linux` cross-compile, and `GOOS=windows go vet ./cmd/client/` is clean.

To finish the acceptance check on macOS with a local broker and relay:

```sh
# 1. Baseline: note the public IP outside the tunnel.
curl -s https://api.ipify.org

# 2. Connect the whole device through the local test relay.
sudo go run ./cmd/client connect --tun \
  -broker http://localhost:8080 \
  -sing-box /opt/homebrew/bin/sing-box

# 3. In another shell, with NO proxy variables set, confirm the exit IP is the
#    relay's and that a utun interface carries the tunnel address.
curl -s https://api.ipify.org
ifconfig | grep -A2 'utun' | grep 172.19.0.1

# 4. Quit with q. The utun device and its routes go away; step 1's IP returns.
curl -s https://api.ipify.org
```

The point of step 3 is that no proxy variable is set: anything answering there proves the routes, not a configured client.

## Docs

`docs/desktop-client.md`'s "Desktop CLI Client" section — which described the deleted linear client and its `sudo go run ./cmd/client connect -broker ...` flow — is replaced with a "Terminal Client" section covering views, keys, proxy vs TUN, the refusal message, the two divergences above, and the headless subcommands. Every command in it was run (except the `sudo` one). The README's "desktop TUN CLI" line, its try-a-client snippet, the repository-layout line, and the docs table follow.

`cmd/client/VERSION` 0.3.0 → 0.4.0. No dependency changes (`golang.org/x/sys` was already a direct dependency and is already in `THIRD_PARTY_NOTICES.md`).


openrung-admin:
All three addressed in 5f84348.

## 1 — TUN readiness could falsely report a working VPN (P1)

Confirmed and fixed; this was the real bug of the three. `172.19.0.1` is inside `172.16.0.0/12`, which is exactly where Docker's default address pool carves bridge networks, so on a host with a matching bridge readiness fired before this sing-box had done anything — and the direct internet probe that follows would then pass over the ordinary network and publish an untunneled session as `CONNECTED`.

Readiness no longer asks "does this address exist somewhere". It asks the kernel **which source address it would send internet-bound traffic from**, and requires that to be the tunnel address:

```go
conn, err := net.Dial("udp4", "1.1.1.1:53")   // route lookup + bind only; no packet is sent
// ready iff conn.LocalAddr().IP == 172.19.0.1 (or the v6 tunnel address)
```

This answers the routing question directly rather than a proxy for it:

- A Docker bridge holding `172.19.0.1` is never the source chosen for an internet destination, so the false positive is gone by construction.
- It also closes a window I had left open: a TUN device can exist with its address assigned *before* sing-box installs routes. The old check passed there; the new one does not, because until the routes land the kernel still answers with the physical interface.
- Another VPN taking the default route correctly reads as not-ready — our tunnel is not carrying traffic.

Both address families are tried, so a v4-only or v6-only host is answered by whichever one has a route. Nothing is sent to either target — connecting a UDP socket is a local route lookup and bind. Verified on this machine: `udp4 1.1.1.1:53 → source=192.168.0.71`, `udp6 → no route to host`, no packets, no privileges.

Because readiness is now a higher bar than binding a socket — the device *and* the routes, plus policy rules on Linux — TUN rungs get `TUNTunnelReadyTimeout` (15s) instead of the 5s proxy window.

Tests: the Docker-bridge regression is pinned directly (`TestTUNReadinessIgnoresAForeignHolderOfTheTunnelAddress`), alongside the not-yet-routed, v6-only, and no-route-at-all cases and the timeout selection.

## 2 — Windows TUN teardown is never graceful (P1)

Confirmed, and it is worse than the report: `CREATE_NO_WINDOW` in `configureSingBoxProcess` means the child has **no console at all**, so `GenerateConsoleCtrlEvent` — the usual substitute for `os.Interrupt` on Windows — cannot reach it either. There is no graceful stop available for this child today without a redesign (job object plus a helper, or sing-box-side IPC), and I have no Windows host to validate one on.

I took the second option you offered: **TUN mode is refused on Windows**, in the CLI's elevation hook, regardless of the process token. The message says teardown is the reason rather than sending the user off to elevate a terminal that would not help. Proxy mode is untouched — it holds no device and no routes, so a forced stop there costs nothing, and it remains fully featured on Windows.

The constraint is now stated where it can be found before someone re-enables it: `connectcore.TUNKillGrace` says the grace is worthless where the interrupt is not delivered, and the `Elevation` interface doc says a host that cannot stop the tunnel gracefully must refuse there. `docs/desktop-client.md` has a "Windows" subsection with the reasoning and the condition for lifting it.

`elevation_windows.go` no longer checks the token at all, so the unused `IsElevated` path is gone with it.

## 3 — Windows UI told users to use sudo (P2)

The wording now comes from a constant owned by the same platform file as the check that enforces it (`tunModeSummary`, `tunModeAdvice`), so the copy and the refusal cannot drift apart. It reaches the Settings Mode row, the mode-change note, the `--tun` flag help, and the usage text.

- Unix: `TUN — whole device (needs sudo)`
- Windows: `TUN — whole device (unavailable on Windows)`, and the note explains that connecting will be refused.

A test asserts the refusal names a fallback, that the Windows path says "not supported on Windows" and never says sudo, and that neither copy string is empty.

## Verification

`go build`, `go vet`, `go test` clean in the root, `desktop/`, and `desktop-volunteer/`; `GOOS=windows go vet ./cmd/client/` and `GOOS=linux` cross-compile clean. The route-source primitive and the non-sudo refusal were exercised live. The `sudo` leg of the acceptance test is still unrun here — sudo needs a password in this environment — and the procedure remains in the PR body; note that step 3 now also exercises the stricter readiness, since the client will not report connected at all unless the routes are actually installed.

openrung-admin:
Fixed in f2da33b. That note was a leftover from the pre-rewrite section that I carried over without rereading it against the Windows gate added in 5f84348.

It now says Windows TUN is disabled pending graceful shutdown support, points at the Windows subsection for the reasoning, and puts the tunnel-driver install checks where they belong — after a graceful stop exists, not instead of one. Swept the rest of the docs and README for the same claim; nothing else asserts Windows TUN support.

PR #136, opened 2026-08-16
Collapse the proxy-port callback seam and its constant cycle

openrung-admin:
Track B cleanup after ADR-001 B3. No UI or connection behavior changes.

## The knot

`internal/proxyconfig` re-exported `Host` and `PortEnv` from `connectcore`:

```go
Host    = connectcore.ProxyHost
PortEnv = connectcore.ProxyPortEnv
```

That import meant `connectcore` could not import `proxyconfig` back, so the engine could not call the resolution policy it needed. Instead it took a callback, `Engine.ResolveProxyPort func() (ProxyPortResolution, error)`, and each host filled it with the same closure translating `proxyconfig.PortResolution` into the identical `connectcore.ProxyPortResolution` — 10 duplicated lines in `cmd/client/host.go` and 10 more in `desktop/vpnservice/service.go`. `ProxyPortResolution` existed only because the type it mirrors could not be named.

## The untangle

`proxyconfig` is now the outright owner of the stable local endpoint — the loopback host, the override variable, the port-validity rule (`ValidPort`, previously duplicated), and the resolution policy — and imports **no openrung package at all**:

```
$ go list -deps openrung/internal/proxyconfig | grep openrung
openrung/internal/proxyconfig

$ go list -deps openrung/internal/connectcore | grep -E 'proxyconfig|clientstate'
openrung/internal/proxyconfig
```

`connectcore` imports it directly, and the cycle is now structurally impossible rather than merely absent — re-adding the back-edge would not compile.

`Engine.ResolveProxyPort` and `connectcore.ProxyPortResolution` are deleted, along with `connectcore.ProxyHost` and `connectcore.ProxyPortEnv`. `LocalProxyPort` calls `proxyconfig.ResolvePort` with the storage hook the engine already had:

```go
resolution, err := proxyconfig.ResolvePort(s.portStore())
```

## Persistence instead of a second hook

Rather than add a new engine field, the existing `Persistence` interface absorbed the two port methods, making it exactly `proxyconfig.PortStore`. That is the property that lets the engine hand its own hook straight to the shared policy, and it is asserted at compile time so it cannot drift:

```go
var _ proxyconfig.PortStore = Persistence(nil)
```

Hosts now wire **one** thing. Both `storeAdapter`s gained two pass-through methods (`*clientstate.Store` already speaks proxyconfig's shape), and both closures are gone. `proxyconfig`'s entry points take small interfaces — `PortStore` for `ResolvePort`, `HelperStore` for `WriteShellHelper` — so it never names the concrete store; a test asserts `*clientstate.Store` satisfies both.

**One trap this introduced, and how it is handled.** Both hosts previously passed a possibly-nil `*clientstate.Store` to `WriteShellHelper`, relying on its `store == nil` check. Through an interface parameter, a nil pointer becomes a *non-nil* interface and would sail past that check into a nil-receiver panic. Each host now has a `helperStore()` that returns a true nil interface when the config directory is unavailable, and each has a test pinning it.

## Behavior preserved

Every invariant has a test; the ones that were previously implicit are now explicit.

| Invariant | Coverage |
| --- | --- |
| Override wins, is not persisted | `TestLocalProxyPortOverrideWinsAndIsNotPersisted` asserts the stored port is left untouched and zero saves occur |
| Automatic selection stable across launches | `TestLocalProxyPortAutomaticSelectionIsStableAcrossEngines`; on the desktop side `TestGetProxyInfoReusesThePersistedEndpointOnTheNextLaunch` over a real `clientstate` dir |
| Stable across processes | `TestLocalProxyPortAdoptsAnotherProcessWinner` — the fake store now models a racing writer, so the loser genuinely adopts the winner's port |
| Persistence failure is a non-fatal warning | `TestLocalProxyPortPersistenceFailureIsANonFatalWarning` covers both no-config-directory and unwritable-store |
| Resolution pinned per process | asserted in the two tests above and in the rewritten `TestLocalProxyPortRetriesAfterResolutionFailure` |
| Transient failure stays retryable | same test, now driven by a real unusable override rather than a stub that returned an error |
| Bind check and actionable error | `TestEnsureProxyPortAvailableReportsStablePortCollision`, unchanged |
| Proxy resolves, TUN never does | `TestTUNConnectSkipsProxyPortAndOSProxy` now asserts zero load *and* zero save calls against persistence |
| Desktop and CLI identical | both wire one `Persistence`; `var _ connectcore.Persistence = storeAdapter{}` in each module |

Two test-quality notes worth flagging:

- `TestGetProxyInfoSurfacesNonFatalPersistenceWarning` set `s.store` but not the engine's hook. Under the old callback that still exercised the unwritable-store path; under the new wiring it would have silently fallen through to the *nil*-store path and kept passing while testing something else. It now wires both through a `withStore` helper that mirrors `Startup`.
- I checked the new desktop stability test is not vacuous by temporarily unwiring `Persistence` in `withStore`: `--- FAIL: TestGetProxyInfoReusesThePersistedEndpointOnTheNextLaunch`.

The connectcore ladder tests dropped their `allocateTestProxyPort` stub and now run the real resolution policy (no store → fresh loopback port + warning), so they exercise the shipped path instead of a fake.

## Verification

`gofmt`, `go vet`, and `go test` clean in the root, `desktop`, and `desktop-volunteer` modules; `GOOS=windows` and `GOOS=linux` cross-compile clean.

### DIFF
--- a/cmd/client/host.go
+++ b/cmd/client/host.go
@@ -13,8 +13,7 @@ import (
 // engineHost bundles the connection engine with the CLI host wiring the TUI
 // and the headless connect driver share: internal/clientstate persistence
 // (recents, crash-recovery proxy snapshot, and the stable proxy port — the
-// same on-disk state the desktop app reads), per-OS system proxy control, and
-// the stable proxy-port resolver (OPENRUNG_PROXY_PORT honored, not persisted).
+// same on-disk state the desktop app reads) and per-OS system proxy control.
 type engineHost struct {
 	engine *connectcore.Engine
 	store  *clientstate.Store // nil when the config directory is unavailable
@@ -33,10 +32,6 @@ func newEngineHost(sink connectcore.EventSink, singBoxPath string) *engineHost {
 	engine.OSProxy = osProxyAdapter{ctrl: proxymode.New()}
 	// TUN mode is gated on this hook; proxy mode never invokes it.
 	engine.Elevation = elevation{}
-	// The one storage hook: recents, the crash-recovery proxy snapshot, and the
-	// stable proxy port the engine resolves through internal/proxyconfig. Left
-	// nil when the config directory is unavailable, which the engine degrades
-	// on rather than fails.
 	if store, err := clientstate.New(); err == nil {
 		host.store = store
 		engine.Persistence = storeAdapter{store: store}
@@ -44,9 +39,8 @@ func newEngineHost(sink connectcore.EventSink, singBoxPath string) *engineHost {
 	return host
 }
 
-// helperStore narrows the store for proxyconfig's shell helper, as a nil
-// interface when the config directory was unavailable: a nil *clientstate.Store
-// inside a non-nil interface would slip past proxyconfig's own nil check.
+// helperStore must return a nil interface, never a nil pointer inside one, which
+// proxyconfig's nil check cannot see through.
 func (h *engineHost) helperStore() proxyconfig.HelperStore {
 	if h.store == nil {
 		return nil
@@ -96,8 +90,6 @@ func (a storeAdapter) SaveRecents(recents []connectcore.RecentNode) error {
 	return a.store.SaveRecents(stored)
 }
 
-// The proxy-port half needs no translation: the store already speaks
-// proxyconfig's shape, which is why the engine can resolve through it directly.
 func (a storeAdapter) LoadProxyPort() (int, bool) { return a.store.LoadProxyPort() }
 
 func (a storeAdapter) LoadOrSaveProxyPort(candidate int) (int, error) {
--- a/cmd/client/host.go
+++ b/cmd/client/host.go
@@ -33,21 +33,25 @@ func newEngineHost(sink connectcore.EventSink, singBoxPath string) *engineHost {
 	engine.OSProxy = osProxyAdapter{ctrl: proxymode.New()}
 	// TUN mode is gated on this hook; proxy mode never invokes it.
 	engine.Elevation = elevation{}
+	// The one storage hook: recents, the crash-recovery proxy snapshot, and the
+	// stable proxy port the engine resolves through internal/proxyconfig. Left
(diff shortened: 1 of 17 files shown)


openrung-admin:
Done in a7138e6 — comments only, no code touched: 135 lines removed, 47 added.

**Kept**, since none of them are inferable from the code:

- loopback-only because the mixed inbound is unauthenticated (`proxyconfig.Host`)
- overrides win for one launch and are never persisted (`proxyconfig.PortEnv`)
- why a nil pointer must reach proxyconfig as a nil interface — stated once as a contract on the interfaces, then a one-line pointer at each of the three conversion sites rather than the reasoning re-derived four times
- why `LoadOrSaveProxyPort` returns another process's winner (on both interface declarations, since a host author reading `connectcore.Persistence` shouldn't have to go find proxyconfig)

**Removed:**

- the package-level paragraph in `proxyconfig.go` about owning the endpoint and the callback seam being gone — that's the PR description, not a code comment
- the "one storage hook: recents, the crash-recovery proxy snapshot, and the stable proxy port" paragraph, which had been pasted into four files
- your example, `// The proxy-port half needs no translation…`, in both copies
- doc comments restating method names (`LoadProxyPort returns the persisted port and whether one was stored`, `SaveProxyEnvScript writes the helper and returns its path`, `ValidPort … exported because the engine applies the same rule`)
- multi-paragraph test preambles, now one sentence of behavior each; e.g. "Automatic selection is stable across launches and across processes: the first launch allocates and persists, and every later engine — in this or another process — is handed the same endpoint by the store" → "Without an override the endpoint is chosen once and reused on later launches"

One fix that isn't cosmetic: the `engineHost` doc still advertised "the stable proxy-port resolver (OPENRUNG_PROXY_PORT honored, not persisted)", which the host no longer has — that behavior moved to proxyconfig, where it is now documented.

`gofmt`, `go vet`, and `go test` clean in all three modules.

PR #137, opened 2026-08-16
Rewrite cmd/relay as a frontend over the relay engine (ADR-001 C2)

openrung-admin:
`cmd/relay/main.go` is now a flag-parsing frontend: it parses flags, maps them onto an `engine.Config`, and renders the engine's status back to the console. All orchestration lives in `internal/relayruntime/engine`.

Deleted from `cmd/relay`: `run`/`runTunnelMode`, `prepareRuntime`/`preparedRuntime`, `register`, `heartbeat`, `heartbeatOrRegister`, `brokerClient`, `cliConfig`+`ApplyDefaults`/`Validate`, `applyFoundationTokenPosture`, `requireDirectModeForFoundation`, `validateWSSRelayConfig`, `applyWSSFronts`, the byte-identical `stopProcess`, the unpinned `hubTLSConfig`, and all of `reachability.go` (`resolveAutoMode`, `probeHTTPClient`). Net −309 lines.

## The CLI surface is a compatibility contract

`deploy/` and the volunteer one-liner consume these; every one still exists with the same name, default, and meaning. `TestFlagSurfaceAndDefaultsAreStable` pins the whole table, and `TestEntrypointFlagsAllExist` re-derives the consumed set from `deploy/relay/entrypoint.sh` itself.

| Consumer | Flags consumed |
|---|---|
| `deploy/relay/entrypoint.sh` | `-mode`, `-xray`, `-config-out`, `-hub`, `-hub-tls`, `-hub-insecure`, `-broker`, `-listen-host`, `-listen-port`, `-heartbeat-interval`, `-connection-log`, `-public-host`, `-public-port`, `-server-name`, `-reality-dest`, `-client-id`, `-reality-private-key`, `-reality-public-key`, `-short-id`, `-wss-fronts`, `-max-sessions`, `-max-mbps` |
| `deploy/relay/volunteer-up.sh` | `-print-label` |
| `deploy/relay/README.md` | `--version` |
| read natively from the environment (never argv, to keep secrets out of `ps`) | `OPENRUNG_VOLUNTEER_TOKEN`, `OPENRUNG_FOUNDATION_TOKEN`, `OPENRUNG_IDENTITY_SEED`, `OPENRUNG_LABEL`, `OPENRUNG_NODE_CLASS`, `OPENRUNG_MODE`, `OPENRUNG_TUNNEL`, `OPENRUNG_HUB_ADDR`, `OPENRUNG_HUB_HTTP_URL`, `OPENRUNG_WSS_FRONTS`, `OPENRUNG_PUNCH_DISABLE` |

`relay -h` is byte-identical to `main`'s apart from the one added flag.

`deploy/relay/foundation-up.sh` and `foundation-wss-host.sh` poll container logs for `registered relay`, so that line is a deployment contract, not decoration — `consoleReporter` re-emits it (and `re-registered relay`, which the same grep matches) from the engine's status stream.

## Engine additions

`ListenHost`, `PublicPort`, and `ConfigPath` — the operator surface that existed only in `cmd/relay`. Each is inert at its zero value, so the desktop volunteer app is unaffected. The listen host resolves through one function shared by the reachability probe's bind and the direct listener, which makes the probe/listener parity invariant structural rather than a comment. The WSS posture gains the listen-host and public-port rules that previously lived only in `cmd/relay`.

## Deliberate upgrades for CLI relays

- **Hub leaf-certificate pinning** — new `-hub-cert-fingerprint` / `OPENRUNG_HUB_CERT_FINGERPRINT`. Off unless a fingerprint is supplied; nothing changes for existing deployments. It is the production-safe alternative to `-hub-insecure` for a hub on a bare IP.
- **`Retry-After`-aware registration backoff** — startup registration retries three times honouring a 429's `Retry-After` instead of failing on the first error.
- **IPv6 rotation rechecks** — a relay whose public address was auto-detected re-checks every 5 minutes and re-registers when the prefix rotates. Only arms when the address actually came from detection; an explicit `-public-host` is pinned and never rechecked.
- **Supervised self-healing** — a session that fails (broker outage, xray crash, hub loss) restarts with capped exponential backoff instead of exiting.

## Resolved divergences

1. **`node_class` and `transport` on the wire.** cmd/relay always sent `node_class: "volunteer"` and `transport: "direct"` explicitly; the engine omitted both. The engine now states them, so **cmd/relay's registration is unchanged** — verified byte-for-byte, identity proof included, by capturing both binaries' register bodies with a fixed identity seed. The desktop volunteer app now sends them too: the broker canonicalizes an absent field to exactly these values (`server.go` `NormalizeNodeClass`, `store.go` `normalizeTransport`) and `relay.IdentityStatement` binds the canonical form either way, so the signed statement and the broker's interpretation are identical.

2. **The process no longer exits on failure.** A relay that cannot register now retries in-process (`PhaseRetrying`, logged) rather than exiting non-zero for the container to restart. Deploy verification is unaffected: it polls for `registered relay` with a timeout, so a relay that never registers still fails the poll. The trade is that a post-registration failure no longer shows up as a rising `RestartCount`; it shows up as `relay stopped, retrying` in the logs.

3. **The connection observer always runs for non-WSS direct mode.** Previously `-connection-log=false` let xray own the public port; now the observer always proxies it (it is the engine's traffic-counter source) and simply has nowhere to write per-connection lines. No records are produced either way, and the privacy posture is unchanged. WSS relays still bypass the observer entirely, so the Hetzner/bunny foundation relays are untouched.

4. **`-skip-xray-run` now binds the public port.** It follows from (3): the observer starts even with xray disabled. The flag is a debug aid and is used by no deploy script.

5. **Loosened `-listen-host dual` validation.** The old rule rejected `dual` alongside `-connection-log=false` or `-skip-xray-run`, because those paths skipped the observer that expands the alias. Both now run the observer, so the rule only survives where it still means something: `RenderXrayConfig` refuses it (xray binds one address), and the WSS listen-host allowlist excludes it. Strictly more permissive, so no existing config breaks.

6. **Fail-fast moved for two flag values.** An invalid label and an out-of-range public port are now rejected at `Start` rather than failing every session. A malformed `-identity-seed` is rejected by the frontend: the engine deliberately regenerates a corrupt seed (a GUI volunteer cannot hand-repair `identity.json`), which for a server would silently churn the relay ID.

7. **Console output.** Engine progress lines and xray's stdout now go to stderr as plain lines alongside the structured log, rather than xray stdout → stdout and relay progress → `slog`. The status-derived lines (`starting relay`, `registered relay`, `re-registered relay`, `relay published via hub`) are still `slog` records with their original fields. Observer lines stay on stdout. On Windows, xray no longer opens a console window (the engine sets `CREATE_NO_WINDOW`).

## Verification

- `go build ./... && go test ./...` clean in the root module, `desktop/`, and `desktop-volunteer/`; `gofmt` clean; `go vet` clean.
- `relay -h` diffed against `main`: identical but for `-hub-cert-fingerprint`.
- Registration request captured from both binaries with a pinned identity: **byte-identical**, signature included.
- End-to-end against a local broker + relay hub, old binary vs new, with matching directory entries (label, public host/port, `transport`, `node_class`, `relay_version`):
  - direct mode — registers, heartbeats, logs `registered relay`
  - tunnel mode — hub publishes the relay, logs `relay published via hub`
  - auto mode — probes, resolves to direct, advertises the probed port
  - `-print-config-only` — identical JSON for both the default and an explicit `-listen-host`
- `deploy/` unmodified except a commented `OPENRUNG_HUB_CERT_FINGERPRINT` entry in `.env.example`. All deploy suites pass: `volunteer-up_test.sh`, `foundation-up_test.sh`, `foundation-wss-host_test.sh`, `ec2-up_test.sh`, `azure-front-door-up_test.sh`, `bunny-wss-front_test.sh`, `hetzner-wss-origin-firewall_test.sh`.

No nested module (`brokerapi/`, `wsscore/`, `punchcore/`) was touched, and no dependency was added.


openrung-admin:
All three confirmed and fixed in 1f57a55.

## 1. Re-registration log suppressed — confirmed, and worse than the report says

The relay ID is not merely "deliberately preserved" — it is *derived* from the identity key broker-side (`internal/broker/store.go:101`, `postgres_store.go:220`), and `prepareIdentity` holds one key for the whole session. So the ID is **always** stable across re-registration and the `re-registered relay` line was unreachable in production, not just usually. My test hid it by simulating `relay_1 → relay_2`, which only the in-memory `fakeBroker` does.

Registration is not derivable from any other `Status` field either — phase stays `Online`, and the endpoint is unchanged. So the engine now carries the signal explicitly:

```go
// Registrations counts successful broker registrations and hub
// publications across the engine's lifetime. It increments even when
// RelayID does not change — the broker derives the ID from the relay's
// identity key, so re-registering after an expired lease returns the same
// one — which is why a caller that reports each registration must watch
// this rather than diffing RelayID.
Registrations uint64 `json:"registrations"`
```

Incremented at the three registration points (initial direct, re-register after an expired lease, hub `OnRegistered`); `consoleReporter` keys off it and still resets on leaving `Online`, so a post-retry session reports as a fresh registration rather than a renewal. Additive for `desktop-volunteer`, which builds its own `State` field by field.

Verified end-to-end against a stub broker that prunes the lease on the second heartbeat and returns a stable ID:

```
=== old ===
INFO registered relay    id=relay_stable_e2e public=203.0.113.7:8443
INFO re-registered relay id=relay_stable_e2e public=203.0.113.7:8443
=== new ===
INFO registered relay    id=relay_stable_e2e label=prune-new public=203.0.113.7:8443
INFO re-registered relay id=relay_stable_e2e label=prune-new public=203.0.113.7:8443
```

New engine test `TestReRegistrationIncrementsTheCounterWithAStableRelayID` pins it: counter 1 → 2 with `RelayID` unchanged. The `consoleReporter` test now drives one stable ID throughout, so it fails without this fix.

*(Visible above and worth noting: the `label` field's source changed. Old logged `desc.Label`, the broker's echo; new logs the engine's configured label — hence the empty value from a stub that doesn't echo it. Against the real broker both print the same thing.)*

## 2. `-mode auto` without `-hub` — confirmed

Restored in `engineConfig()` rather than the engine, which intends the hubless degrade for a GUI with no hub configured yet. `normalizeMode` only yields `auto` for an explicit request or a configured hub, so this catches exactly the old case. Same message, same exit code:

```
old: ERROR invalid relay config error="hub is required in auto mode for reachability probing (set -hub or use -mode direct)"  exit=2
new: ERROR invalid relay config error="hub is required in auto mode for reachability probing (set -hub or use -mode direct)"  exit=2
```

## 3. Explicit `-public-port 0` — confirmed

Rejected at the CLI boundary; zero stays the engine's "advertise the listen port" default:

```
old: ERROR invalid relay config error="public-port must be between 1 and 65535"  exit=2
new: ERROR invalid relay config error="public-port must be between 1 and 65535"  exit=2
```

One deliberate difference: the old rule lived in the `direct` branch only, so `-public-port 0 -mode tunnel` was accepted and ignored. The new check is unconditional — the flag defaults to 443, so an explicit 0 is never meaningful in any mode.

## Re-verified

Build, vet, test, and gofmt clean in the root module, `desktop/`, and `desktop-volunteer/`. All deploy suites still pass. The registration wire is untouched by this commit — `Registrations` is engine-internal status, not a request field.

openrung-admin:
Both confirmed and fixed in 687c3ae. Reproduced each against the base binary first:

```
old  foundation+auto no hub : exit=0     new (1f57a55)  exit=2   ← regression
old  tunnel -public-port 0  : exit=0     new (1f57a55)  exit=2   ← regression
```

## Root cause

Both rules read `Config.Mode`, which is not the mode the relay runs in. The engine's `applyFoundationTokenPosture` overrides it to direct so the credential can never reach a hub — so `-foundation-token -mode auto` never probes, and judging it as auto was wrong. The same blind spot made me apply the public-port rule in every mode when cmd/relay only ever applied it in `direct`.

The tempting fix is to re-derive "a foundation token forces direct" in the frontend, which is exactly the duplication this PR exists to delete. So the engine now answers the question:

```go
// EffectiveMode reports the mode this config will actually run in, which is
// not always Mode: a foundation token forces direct, so that its credential
// can never reach a hub. A caller layering its own rules on the mode — the CLI
// rejects a hubless auto config that the engine would instead degrade to
// direct — must consult this, or it will judge a mode the posture is about to
// override. A config whose posture is self-contradictory has no effective mode
// and returns "", leaving that error for Start to report in full.
func (c Config) EffectiveMode() string
```

The two CLI-only rules are now a per-mode switch keyed off it, the same shape cmd/relay's own `Validate` had:

```go
switch cfg.EffectiveMode() {
case engine.ModeAuto:
	if f.hubAddr == "" { ... }        // nothing to probe against
case engine.ModeDirect:
	if f.publicPort < 1 || ... { ... } // only direct advertises the port
}
```

The `""` case matters: `-foundation-token fnd -node-class volunteer` is contradictory, so neither rule fires and `Start` reports the actual node-class conflict rather than a misleading mode error. `TestModeSpecificRulesRespectTheEffectiveMode` pins that, along with the fact that a token forcing direct brings the public-port rule *with* it.

## Parity re-verified, all four cases

```
### previously-accepted (must stay accepted)      ### originally-reported P2s (must stay rejected)
old  foundation+auto no hub : exit=0              old  auto no hub           : exit=2
new  foundation+auto no hub : exit=0              new  auto no hub           : exit=2
old  tunnel -public-port 0  : exit=0              old  direct -public-port 0 : exit=2
new  tunnel -public-port 0  : exit=0              new  direct -public-port 0 : exit=2
```

This also withdraws the deliberate tightening I flagged in my last comment — you were right that a PR whose contract is "no new rejections" shouldn't be quietly adding one. `-public-port 0` in tunnel mode is accepted and ignored again, exactly as before.

New `TestEffectiveModeReportsThePostureOverride` covers the engine side. Build, vet, test, and gofmt clean across the root module, `desktop/`, and `desktop-volunteer/`; all deploy suites pass. The registration wire is untouched by this commit.

PR #138, opened 2026-08-16
Golden contract vectors + fix the Go classifier's Winsock gap

openrung-admin:
The failure-classification token set, the relay-descriptor decode, and the broker-front list are load-bearing contracts across four clients in two repos, enforced until now only by convention. This adds `contract/vectors/` as the machine-checked source of truth, runs it from the Go suites, and fixes the Windows gap it exposes.

## The Winsock fix

`internal/clienttelemetry/classify.go` matched only Go's `syscall.E*` names. On Windows those are invented `APPLICATION_ERROR` values (1<<29 and up) while the net package surfaces the raw Winsock numbers from `connectex`/`WSARecv` untranslated — so the switch compiled cleanly there and never fired, and **every Windows socket failure degraded to `unknown`**.

The errno switch now pairs each POSIX errno with its Winsock number (10061/10054/10051/10065), and `WSAETIMEDOUT` (10060) is named in the timeout rule, after the DNS check — `syscall.Errno.Timeout` on Windows reports true only for Go's invented `EAGAIN`/`EWOULDBLOCK`/`ETIMEDOUT`, so a real Winsock timeout reached neither existing timeout check. These are the same conditions wsscore's build-tagged classifier maps (`wsscore/failure_windows.go`).

The numbers are matched unconditionally rather than behind a build tag: they sit far above every errno a POSIX kernel returns (Linux tops out at 133, darwin at 106), so they cannot collide with a real one, and matching them everywhere is what lets CI exercise the Windows rows on Linux and macOS. Verified: the five Winsock rows fail against the pre-fix classifier and pass against the fix, on darwin.

```
--- FAIL: TestClassificationVectors/errno_wsaeconnrefused
    ClassifyError(errno 10061) = "unknown", want "connection_refused"
--- FAIL: TestClassificationVectors/errno_wsaeconnreset
    ClassifyError(errno 10054) = "unknown", want "connection_reset"
--- FAIL: TestClassificationVectors/errno_wsaenetunreach
    ClassifyError(errno 10051) = "unknown", want "network_unreachable"
--- FAIL: TestClassificationVectors/errno_wsaehostunreach
    ClassifyError(errno 10065) = "unknown", want "network_unreachable"
--- FAIL: TestClassificationVectors/errno_wsaetimedout
    ClassifyError(errno 10060) = "unknown", want "timeout"
```

## The vectors

`contract/vectors/`, each file carrying its own doc header naming its consumers and the version-bump rule:

- **`classification.json`** — 69 rows covering every branch of the ladder (cancellation → WSS taxonomy → selection sentinels → HTTP status → errno → DNS → TLS → permission → process exit → timeout → unknown), including the Winsock codes. Inputs are described abstractly by `kind` so each language builds its own native error; `kinds` says which suites can construct each kind, and a row may narrow that with its own `suites` plus a note saying why. Two coverage tests check the vectors themselves: every token of the closed enum is produced by some row, and no row expects a token outside it.
- **`relay_decode.json`** — three signed directory bodies (a normal API page, a filter-exercising mix, a forward-compatible body from a newer broker) with expected descriptor fields, usability probes at several instants, and seven wire-schema rejections.
- **`broker_fronts.json`** — the advertised front list, the two-phase race order, the hostname-shape classification that assigns a front to a phase, and the discovery order built from a configured primary.

Go suites: `contract/relay_decode_test.go` and `contract/broker_fronts_test.go` (against brokerapi's fetch path, `internal/relay`, and `internal/client`'s selector), plus `internal/clienttelemetry/contract_vectors_test.go`. Each pins the vector-file version it was written against in a constant, so an edited row fails the build here while the mobile repo's check script catches the un-re-vendored copy there. No nested module is touched.

The follow-up PR in `openrung-mobile-app` vendors these files with that check script and adds the JUnit/XCTest/Jest suites.

## Resolved divergences

- **Winsock text vs codes.** The prompt asked for Winsock error *text* as well as codes. Classification is by numeric code only: Windows localizes its socket error strings through `FormatMessage`, so text matching would silently stop working outside an English install. The vector file states this in its `windows_note`.
- **`WSAECONNABORTED` (10053) left unmapped.** Its closest POSIX analogue is `EPIPE`, which this ladder classifies as `unknown` (there is no broken-pipe token); wsscore maps both to a reset because its taxonomy has phase-split reset tokens. Mapping the Winsock code alone would make the same broken connection report different tokens per platform — the exact skew this PR removes. Both codes are pinned as `unknown` rows with notes, so the gap is visible rather than forgotten.
- **No signature on the decode vectors.** The production relay-list signing key is offline, so no vector can be validly signed. The bodies carry the in-body signing envelope (`not_after`, `key_id`, `channel`, `limit`) because those fields are what a client reads, and the Go suite drives them through brokerapi's real fetch path over loopback — the development exception that validates the wire schema without a signature. Signature verification stays brokerapi's own concern, covered by its tests.
- **No `relay_version`/`volunteer_version` precedence case.** A body carrying only `relay_version` is rejected outright by the wire schema, so the two keys never disagree in practice and the TypeScript model — which exposes only the legacy name — cannot diverge. Pinned as the invalid case `relay_version_without_alias` instead, which is the more useful statement: dropping the alias is a breaking wire change, not a decoder change.

## Verification

`go build ./...` and `go test ./...` clean in the root module and in `desktop/` and `desktop-volunteer/`; `go vet ./...` clean; gofmt clean; `GOOS=windows` and `GOOS=linux` build and vet clean, so the Windows-side constructs compile where they ship.


openrung-admin:
Fixed in 84b11f8 — you're right on both counts, and the second one is the more important.

**WSAEACCES (10013) now maps to `permission_denied`.** Matched at the permission rung rather than in the errno switch, so it lands where its POSIX counterparts do and stays ordered after the DNS and TLS rules. `internal/relayruntime/probe_bind_errors_windows.go` documents this exact trap for its bind probes, so the ladder was the odd one out.

**The pairing convention is now a check, not a convention.** That is the real defect here: the file's own layout — every POSIX errno beside its Winsock number — made 10013's absence read as deliberate when it was an oversight, and nothing would have caught the next one. Every errno row now names the row on the other platform describing the same condition, and `TestClassificationVectorsErrnoPairing` verifies the pair exists, sits on the other platform, and expects the same token, plus that neither platform reports a token the other cannot.

Both guards fail on the pre-fix state, from either direction:

```
# vectors without the WSAEACCES row (the state you flagged)
--- FAIL: TestClassificationVectorsErrnoPairing
    errno vector "errno_eacces" pairs with "errno_wsaeaccess", which does not exist
    errno vector "errno_eperm" pairs with "errno_wsaeaccess", which does not exist
    a POSIX errno reports "permission_denied" but no Winsock code does; either map
    its counterpart or pin the gap with a row

# classifier without the WSAEACCES mapping
--- FAIL: TestClassificationVectors/errno_wsaeaccess
    ClassifyError(errno 10013) = "unknown", want "permission_denied"
```

Pairing is many-to-one where the platforms disagree on granularity: `EACCES` and `EPERM` both pair with `WSAEACCES`. `EPIPE`/`WSAECONNABORTED` stay paired on `unknown`, which is what makes that gap a decision rather than an omission — the same standing this one lacked.

`classification.json` stays at `version 1`: it has not merged or been vendored anywhere yet, and adding a row is the additive case the README already exempts from a bump. Re-verified clean: `go build`/`go vet`/`go test -race ./...` in the root module and both desktop modules, `GOOS=windows` vet clean, gofmt clean.

openrung-admin:
Fixed in bc42741. You're right, and the harness was worse than "unverifiable" — it was fail-unsafe.

**The rule existed but was never named.** `relay.NormalizeNodeClass` has exactly two callers, both ingest-side: `internal/broker/server.go:178` (registration) and `internal/relayruntime/engine/engine.go:312` (the relay's own config). The harness was the third and the only read-side one, which is the bug. The actual client-side rule is strict equality against `NodeClassFoundation`, implemented ad hoc at two sites — `cmd/client/views.go` (which already carried it as the comment *"A missing class is the volunteer class, per the descriptor contract"*) and `internal/connectcore/wss.go`'s WSS gate.

So: named it `relay.EffectiveNodeClass`, and wired it into both sites. That substitution is exactly behavior-preserving — it is the same strict equality by construction — and it means the harness now checks the same function the client runs, rather than a test-only stand-in that could drift from the two real call sites.

**Added the rows the file demanded**: a future class name (`partner`) and a case variant (`Foundation`), both decoded verbatim, both read as volunteer, and both still usable — the class is provenance, not a usability condition, and dropping such a relay would strand a client on a directory it merely does not fully understand.

Confirming your prediction against the old harness with the new rows present:

```
--- FAIL: TestRelayDecodeVectors/forward_compatible
    relay 1 node_class "partner": node_class must be "volunteer" or "foundation"
    relay 2 effective node_class = "foundation", want "volunteer"
```

The first line is the failure you called. The second is the one I had not seen: `NormalizeNodeClass` trims and lowercases, so the old harness asserted that `"Foundation"` **is** the foundation class — fail-unsafe, on the class that gates the WSS transport. A convenience that is correct for operator input at ingest is a vulnerability when reading a value that has already crossed the trust boundary. `EffectiveNodeClass` degrades in one direction only, and `TestEffectiveNodeClass` pins the case variants, the whitespace variants, and the unknown name.

`expectation_note` now states the strictness and the direction explicitly, and warns off implementing the rule with an ingest-side validator. Still `version 1` — additive rows, nothing vendored yet. Re-verified: build/vet/`test -race ./...` in the root module and both desktop modules, `GOOS=windows` build clean, gofmt clean.

openrung-admin:
Fixed in bab5243. Copy-paste from `relay_decode.json`, exactly as you read it — and checking the claim turned up a second one in `broker_fronts.json`.

**`classification.json` has three suites, not four.** There is no TypeScript classifier for a Jest suite to run these rows against — `grep -rln "failure_reason\|failureReason\|classify" --include="*.ts" src __tests__` in openrung-mobile-app returns nothing. The React Native layer never sees a connect failure: Android's VPN service, iOS's PacketTunnel, and native brokerapi own that path and report the token themselves. So adding the promised suite would have meant writing a fourth classifier, not wiring up an existing one. `suites` is now `["go", "kotlin", "swift"]`, with a `no_typescript_note` recording why the asymmetry is deliberate — the thing the boilerplate obscured.

**`broker_fronts.json` overclaimed in the other direction.** Its header implied the mobile suites check the phase ordering; they do not implement it. They hand native brokerapi a single primary (`AppConfig.defaultBrokerURL` / `DEFAULT_BROKER_URL`) and Go owns the racing — their own comments say so. Their real claim is that the fronts they hardcode still appear in `default_order`, which is what catches a front rotated in one repo and not the other. The header now says that, says they must *not* assert the phases, and adds a `scope_note` keeping `UPDATE_MANIFEST_URLS` (a different channel with its own check) out of it.

**The claim is now structured data with a check.** Each file declares `suites`, and `TestClassificationVectorsSuiteDeclaration` fails when a declared suite is claimed by no row, or a row is claimed by an undeclared suite:

```
# declaring the "ts" suite the old header claimed
--- FAIL: TestClassificationVectorsSuiteDeclaration
    suite "ts" is declared as a consumer but no row is claimed by it — it would
    iterate this file, assert nothing, and pass vacuously; either give it rows or
    drop it from "suites"
```

That is the failure mode worth guarding: a header claim was unfalsifiable prose, so the vacuous pass looked like coverage. The mobile check script now also has a machine-readable answer for which suites each file requires.

README reconciled — the table carries a Suites column sourced from the same field, and the prose says explicitly that not every file has every suite and why. Still `version 1`, nothing vendored yet. Build/vet/`test -race ./...` clean in the root module and both desktop modules; all three files re-validated as JSON; gofmt clean.

openrung-admin:
All four fixed in 7821716.

**1. `suites` was enforced in one file out of three.** Exactly the inert-claim problem the field was added to solve — `broker_fronts_test.go` did not unmarshal it at all. The shared checks now live in the `contract` package and run over every file: version present, `suites` non-empty, no duplicates, only known identifiers, and agreement with whether a Go suite really runs that file (kept as an independent list, so the two statements can disagree and be caught). Per-row claiming stays with the classification vectors, whose rows are claimed individually; the other two are consumed whole, so what is checkable is the declaration's well-formedness — I would rather state that limit than imply parity. Also added: `relay_decode`'s `invalid.suites` must be a subset of the file's `suites`, and a new file in `vectors/` must join the checks or `TestVectorFilesAreDeclared` fails.

**2. Pairing was evadable at token granularity.** Correct, and your example is the exact one — a new `ENETDOWN` row pointing at `errno_wsaenetunreach` satisfied both the pair check and the token-set check while `WSAENETDOWN` went unmapped, because the token was already covered by the pair it borrowed. Pairing is now one-to-one by default, and a collapse must be declared in `errno_pair_exceptions` with a reason:

```
--- FAIL: TestClassificationVectorsErrnoPairing
    errno vector "errno_enetdown" pairs with "errno_wsaenetunreach", which pairs back
    with "errno_enetunreach" instead — declare the collapse in errno_pair_exceptions
    with a reason, or give "errno_enetdown" its own counterpart
```

The two genuine collapses are declared: `EPERM` (Winsock has no distinct socket EPERM) and the wrapped `ECONNREFUSED` variant (not a distinct condition). A reason is required, since a collapse without one is indistinguishable from a forgotten counterpart. This does not make the collapse itself provable — nothing can mechanically know Windows has a distinct code for a condition — but it moves the claim from one inconspicuous field on a new row to a visible edit of a shared table.

**3. Bump rule vs. this PR's history.** Fair — I justified each in-PR edit in a comment rather than in the rule. The README now says the version binds from the first vendoring, because it coordinates copies and there is none before that; within the PR that introduces a file, rows are drafted in place at the version they will land on.

**4. `errno_wsaeaccess` → `errno_wsaeacces`.** Doubled S in the row id; the `symbol` field was already right.

Build/vet/`test -race ./...` clean in the root module and both desktop modules, `GOOS=windows` build clean, all three files re-validated as JSON, gofmt clean.

PR #139, opened 2026-08-17
Share the Winsock table, break the token-list cycle, machine-enforce vector bumps

openrung-admin:
Closes the four structural findings deferred from PR #138's review. Wire behavior and the classification ladder's observable output are unchanged: every existing row in `contract/vectors/` passes with identical tokens (`classification.json` is untouched; `relay_decode.json` changed only its Go-only invalid-case `reason` spelling, see item 5).

## 1. One Winsock errno table, shared through wsscore

`wsscore/winsock.go` (untagged) is now the single definition of every Winsock number the repo matches: exported `WSAE*` constants plus an enumerable `WinsockErrnos()` table. `internal/clienttelemetry/classify.go`'s hand-copied const block is gone; both classifiers consume the shared constants but keep their own errno→token mapping — the table shares numbers and names, never token decisions. The const-block comment that wrongly claimed wsscore "maps the same conditions to the same tokens" is replaced by one stating the real relationship and naming 10053 as deliberate.

`TestWinsockTableDivergence` (internal/clienttelemetry) walks the shared table and asserts per code that `ClassifyError` and `wsscore.SocketErrnoReason` agree, or that the divergence is declared with the vector row pinning it:

| symbol | client | wsscore | pinned by |
| --- | --- | --- | --- |
| WSAECONNABORTED (10053) | `unknown` | `connection_reset` | `errno_wsaeconnaborted` |
| WSAETIMEDOUT | `timeout` | `tcp_timeout` | `errno_wsaetimedout` |
| WSAEACCES | `permission_denied` | `unclassified` | `errno_wsaeacces` |

**The posix/windows split is collapsed**, not kept: `failure_posix.go` / `failure_windows.go` / `failure_windows_test.go` are deleted and `classifyDialFailure` matches both the portable `syscall.E*` names and the raw Winsock numbers unconditionally, exactly as classify.go already did — the no-collision argument is identical. wsscore's Windows mappings are now exercised on Linux/macOS: `TestClassifyWinsockErrnos` walks the whole table, and the precedence test gained phase-variant Winsock rows. A range check (`10000–11999`) replaces the old windows-tagged pin test: a reversion to portable names leaves that range on every platform, so it fails on any host instead of needing a Windows run.

## 2. The token-coverage circle is broken

The allowlist is now the package-level `wssReasonAllowlist` slice (plus an explicit, currently empty `wssReasonExcluded` map), and the four hand-synced copies reduce to one source plus machine-checked projections:

- `TestWSSReasonAllowlistMatchesVectors`: (a) the allowlist equals the set of expect values of the vectors' pass-through `wss_reason` rows; (b) every member is in `vectors.Tokens`.
- `TestWSSReasonAllowlistCoversWsscore`: (c) every token in wsscore's new machine-readable registry `Reasons()` is in the allowlist or in `wssReasonExcluded` (never both, never neither), and every allowlist member still exists in the registry.
- `TestWSSFailureReasonAllowlist` now iterates the real slice instead of its own copy.
- In wsscore, source-parsing tests (`registry_test.go`) keep `Reasons()` and `WinsockErrnos()` in lockstep with their const blocks, since Go cannot enumerate constants at runtime.

Demonstration (scratch, not committed) — adding `ReasonFakeToken = "fake_token"` to wsscore's const block:

1. constant only → wsscore `TestReasonsCoversTheConstBlock`: `constant ReasonFakeToken ("fake_token") is missing from Reasons()`
2. \+ registry entry → `TestWSSReasonAllowlistCoversWsscore`: `wsscore token "fake_token" is neither in the allowlist nor in wssReasonExcluded`
3. \+ allowlist entry → `TestWSSReasonAllowlistMatchesVectors`: `no wss_reason vector row pins it` and `the vectors' closed token list does not contain it`

So a token added anywhere must produce a vector row (reaching every suite via the version bump) or a named exclusion, or CI fails.

## 3. Machine-enforced vector version rule

New `contract-vectors-version` job in go-checks.yml (hardened in the second commit after review): any change to a vectors file must carry a top-level `"version"` that is a positive integer **strictly greater** than the previous one — a downgrade or reused number would make one version name two contents across history. The diff is taken with `--name-status -z -M`, so a **rename is an edit of the old file** requiring a bump, **deletion is a hard failure** (until the mobile vendoring check exists), a **non-ASCII filename** cannot be C-quoted into the wrong arm by `core.quotepath`, and unhandled diff statuses fail closed. The pathspec (`'*contract/vectors/*.json'`) matches the vectors directory anywhere in the tree, so the planned ADR-001 D3 move cannot silently disable the job. It runs on PRs **and on every push to main**: with no up-to-date-branch protection, two independently green PRs can race the same bump and merge cleanly (the #75 failure mode) — the push run judges exactly what landed and turns main red. A file new in the PR passes at the version it lands on (the README's drafting rule).

Manual test, run with the job's script extracted verbatim against committed diffs:

- this PR's real diff (relay_decode.json 1→2): **pass** (`version 1 -> 2`)
- edit without a bump: **fail**, naming the file
- downgrade 2→1 with row edits: **fail**
- rename+edit at the same version: **fail**; rename with a bump: **pass**
- outright deletion: **fail**
- `café.json` edited without a bump: **fail**, correctly attributed (not routed to the deleted-file arm)
- move to `connectcore/contract/vectors/` without a bump: **fail** (pathspec still matches, and the rename crosses directories)
- bump 1→3: **pass** (strictly greater, gaps allowed)
- push-mode (oldrev = direct parent): second racer landing row changes at the already-bumped version: **fail**
- new file: **skipped** (`new at version 1; the bump rule binds from its next edit`)

contract/README.md's "Changing a vector" section documents the enforcement.

## 4. One version gate: `contract.LoadVersioned`

`LoadVersioned(name, wantVersion, target)` loads, validates the version (positive + equal to the suite's pin), and carries the single copy of the mismatch guidance — `grep -rn "this suite was written against"` matches only inside it. The three suites keep their pin constants and lose their triplicated blocks; `vectors_test.go`'s `vectorHeader` drops its `Version` field and positivity check, which moved into `LoadVersioned`, so version-validation logic exists exactly once.

## 5. (Optional item, taken) Stable reason codes for relay-decode invalid cases

The invalid cases' `reason` values were exact substrings of brokerapi's Go error messages inside the vendored file, so a Go-internal reword would have forced cross-repo churn. They are now stable codes (`relay_list_missing_required_field`, `relay_missing_required_field`, `wss_front_missing_required_field`, `unknown_field`); the code→wording translation lives in the non-vendored `relayInvalidReasonMessages` fixture in `contract/relay_decode_test.go`. The invalid cases are declared `"suites": ["go"]` — no mobile suite reads them — so the mobile-facing schema is undisturbed. Per the rule item 3 enforces (which runs on this very PR), `relay_decode.json` is bumped to version 2 with its pin.

## Resolved divergences

- **wsscore's errno matching became unconditional.** On POSIX the classifier now also matches raw Winsock numbers (previously → `unclassified`); on Windows it now also matches the portable `syscall.E*` names (Go's invented values). Neither input can be produced by that platform's real network stack — POSIX errnos top out far below 10000, and the invented Windows values sit at 1<<29 and up — so observable output is unchanged for every reachable input; the gain is CI-testability of the Windows mappings.
- **wsscore's explicit timeout case briefly dropped `ETIMEDOUT`** between the first and third commits; review caught that `errors.As` binds the shallowest `net.Error`, so a custom wrapper reporting `Timeout() == false` shadows a wrapped POSIX `ETIMEDOUT` even though the errno was extracted — the old build-tagged table's "redundant" comment was wrong about that edge. The explicit match is restored (`errno == syscall.ETIMEDOUT || errno == WSAETIMEDOUT`) and regression-tested with a false-Timeout wrapper, so the collapse is now genuinely behavior-preserving on both platforms.
- **classify.go's timeout rung gained the same explicit `ETIMEDOUT` match** (third round): the shadowing that review found in wsscore applied identically here — a `net.Error` wrapper reporting `Timeout() == false` around a POSIX `ETIMEDOUT` classified as `unknown` on this ladder while wsscore reported its timeout family, so the platform chose the token. It now reports `timeout`. No vector row pinned the old behavior (the wrapper shape is not expressible in the vectors' input schema, and is Go-specific); a Go regression test covers it.
- **classify.go's comment claimed the two classifiers map "the same conditions to the same tokens" — they never did.** The rewritten comment states the actual relationship (shared numbers, independent mappings) and names 10053; the divergence test makes the claim machine-checked.
- **`relay_decode.json`'s `reason` semantics changed** from Go-message substring to stable code (item 5), with the version bump and a rewritten `invalid.note`. The mobile repo has not vendored yet, so no copy coordination was needed beyond the bump itself.
- **The README's "binds from the first vendoring" rule is now stricter in practice:** the CI job binds from merge onward (it cannot see vendoring state), exempting only files new in the PR. The README paragraph was rewritten to say so.
- **`vectorHeader`'s version-positivity check moved into `LoadVersioned`** so version validation has one home; the header test now checks only the suites declaration.

## Verification

- `go build ./...` + `go test ./...` green in the root module, `wsscore/`, `desktop/`, and `desktop-volunteer/`; `go vet` green in root and wsscore; gofmt clean.
- `wsscore/` builds standalone; `wsscore/VERSION` bumped 0.5.0 → 0.6.0 (new exported API: `WSAE*`, `WinsockErrnos`, `Reasons`, `SocketErrnoReason`); `wsscore/v0.6.0` is untagged on origin.
- `grep -rn` for each of 10013/10051/10053/10054/10060/10061/10065 finds exactly one Go definition repo-wide (winsock.go); remaining hits are the comment naming 10053 as deliberate, the divergence test's comment, and the vector rows themselves.
- brokerapi/ and punchcore/ untouched; registration, heartbeat, directory, and telemetry wire paths untouched.


openrung-admin:
Review findings on the `contract-vectors-version` gate: all five fixed in 68b4e1b, and the PR body's item-3 section is updated with the full verification matrix.

- **Downgrade/reuse** → the new version must now be strictly greater than the old (`[ "$new" -gt "$old" ]`); the 2→1 repro fails with exit 1.
- **Rename+edit / deletion bypass** → the diff is now `--name-status -z -M`: `R*` is treated as an edit of the old file requiring a bump (verified failing at the base version, passing with a bump), `D` is a hard failure until the mobile vendoring check exists, and any unhandled status fails closed.
- **Racing green PRs** → the job also runs on pushes to main (the workflow already triggers there), diffing `github.event.before..HEAD`, so the second racer's merge — rows changed at the already-bumped version — turns main red. Verified with oldrev set to the direct parent.
- **quotepath** → NUL-delimited parsing; the `café.json` repro is now correctly attributed and fails instead of exiting green through the deleted-file arm.
- **D3 self-disable** → pathspec is `'*contract/vectors/*.json'`, which matches the directory anywhere in the tree; verified against a move to `connectcore/contract/vectors/` (fails without a bump, both because the pathspec still matches and because the move is a rename requiring one).

contract/README.md's enforcement paragraph now states the strict-monotonic rule, the rename/deletion handling, and the push-to-main run.

openrung-admin:
Second round of findings: all three fixed in ce8f7d1.

- **wsscore no longer self-pinned exact numbers** → `TestWinsockTableValues` pins every shared-table symbol to its documented Winsock number inside the module. Verified by scratch repro: swapping 10051↔10065 in winsock.go fails the test naming both symbols, where the range/distinctness checks (and every classification test, since both map to `network_unreachable` today) stayed green. The exact numbers now appear twice in wsscore — once as the definition, once as this test's independent restatement of the documented values — which is the point.
- **POSIX ETIMEDOUT belt** (plausible → confirmed) → the finding is right and the old "redundant on POSIX" comment was wrong: `errors.As` binds the shallowest `net.Error`, so a wrapper with `Timeout() == false` shadows a wrapped `ETIMEDOUT`. The explicit `errno == syscall.ETIMEDOUT` match is restored alongside `WSAETIMEDOUT`, with a `falseTimeoutWrapper` regression case in the precedence table — it classifies as `unclassified` without the fix, `tcp_timeout` with it. The "Resolved divergences" bullet in the PR body that claimed the drop was behaviorally identical is corrected.
- **Dead `Version` field** → deleted. It was actually dead on all three vector structs, not just relay-decode's; all three are cleaned.

Root, wsscore, desktop, and desktop-volunteer all build and pass with `-count=1`; gofmt clean.

openrung-admin:
Third round: verified all three findings against the code before fixing — all real — and fixed in 7c15006.

- **ETIMEDOUT on one ladder only** → confirmed by live repro: `ClassifyError(falseTimeoutWrapper{ETIMEDOUT})` returned `unknown` while `wsscore.SocketErrnoReason`'s rule (post-fix) keeps the timeout family — same condition, platform-determined token. classify.go's timeout rung now names `syscall.ETIMEDOUT || wsscore.WSAETIMEDOUT`, mirroring wsscore, with `TestClassifyShadowedTimeoutErrnos` covering both errnos behind a false-Timeout wrapper. Collateral checked: no vector row pins the old `unknown` (the wrapper shape isn't expressible in the vectors' input schema and is Go-specific), every existing vector still passes, and all higher rungs (DNS, TLS, permission, exec) still win over it. Noted under Resolved divergences in the body, since Kotlin/Swift mirror this ladder.
- **Fail-open `old=0`** → confirmed by repro: a base file carrying `"version": "not-a-number"` edited to version 1 printed a fabricated `version 0 -> 1` and exited 0. The fallback is now a hard failure naming the file (`refusing to fail open`); re-verified failing on the same repro and still passing on this PR's real diff.
- **Sibling version jobs PR-only** → confirmed (all three had `if: pull_request` while `*-tag.yml` fires only on VERSION changes, so a second racing merge ships module code its tag doesn't name). punchcore/wsscore/brokerapi-version now run on pushes to main with the changed-without-bump check; the `git ls-remote` tag-freshness probe stays PR-only because on a push the tag workflow may be creating that very tag concurrently, and probing would race it into a false failure. Verified with the extracted script: PR mode passes on this PR (0.5.0→0.6.0, tag fresh), push mode fails the racer's merge (module diff, VERSION unchanged vs parent), push mode passes a legit merge without probing, and an unreachable before-SHA skips cleanly.

All four modules build and pass with -count=1; gofmt clean; workflow YAML parses.

PR #141, opened 2026-08-18
Geographic diversity slots in the broker's ranked relay page

openrung-admin:
## Mechanism

Iranian clients pile onto the Seoul relays (transfer overage) while the Hetzner EU relays idle at 1–3% of allowance, because the broker's single global ranking can serve a page with no EU relay on it. This PR adds **diversity slots + the already-shipped client latency bucketing = de facto geo steering with zero geo code**: after the global sort, if a coarse region present in the eligible fleet (asia / europe / americas / oceania, from a small static country-code map covering the fleet's deploy-target countries) is missing from the page head, the region's best-ranked below-fold relay swaps with a relay from the page tail — at most 2 slots, never the global #1, deterministic, page size unchanged. Every client already probes the whole served page (`RelayRankMaxProbes = 8` ≥ the default page of 5) and reorders by 30 ms latency bucket, so from Iran an EU relay (~110 ms) decisively beats Seoul (~270 ms) while from China Korea (~60 ms) still wins — each client steers itself, and the broker never learns or guesses where the client is.

## Behavior & invariants

- Runs at page assembly in both list handlers (`/api/v1/relays` and the mirror channel), strictly after the global sort and after `store.List`'s eligibility filtering; the signed envelope, page size, and echoed limit are untouched.
- Global #1 never displaced; at most 2 tail slots repurposed; unknown/unmapped country codes can neither win a slot nor count a region as represented, and `resolveRelayGeo` warns when a relay's country falls outside the region map so the map cannot silently fall behind the fleet.
- If every fleet region is already on the page (or the fleet fits in the page), the served bytes are identical to today's.
- A fill never evicts a region's only page representative — coverage can only grow, never trade one missing region for another.
- Displaced relays swap below the fold, and the diversity pass runs **before** `reserveWSSCandidate` and keeps its fills out of the slot the reservation will claim — the page never loses its WSS fallback to a diversity fill, and the reservation never cancels a fill.
- Applies only to the default `global` ranking mode; `legacy` is served untouched.
- Rollback: `OPENRUNG_RELAY_PAGE_DIVERSITY=false` (default enabled, state logged on the startup line, documented in deploy/broker/README.md and .env.example). The flag fails closed: an unrecognized value refuses to start rather than silently keeping the feature running. No client changes, no GeoIP, no new ranking mode.

## Tests

Table-driven tests in `internal/broker/diversity_test.go`: already-diverse page unchanged, KR-heavy head + below-fold EU relay promoted into the tail with #1 untouched and length unchanged, two missing regions fill exactly two slots, three missing regions capped at two, sole-representative protection, all-slots-protected no-op, fleet smaller than page unchanged, unknown country codes never win a slot, single-slot page never displaces #1, determinism across repeated runs, and `TestDiversityFillSurvivesWSSReservation` composing both page passes on the live fleet's shape; plus handler-level tests through `NewServer` for the default-on behavior, the disabled flag (exactly today's output), and legacy mode, and fail-closed parsing tests for the rollback flag in `cmd/broker`.

## Acceptance after deploy

`connection_succeeded` telemetry already records `brokerIndex` and the client's probe latency, so IR sessions shifting toward EU relays is observable with no new instrumentation.

### DIFF
--- a/cmd/broker/main.go
+++ b/cmd/broker/main.go
@@ -67,6 +67,9 @@ func run() error {
 	if err != nil {
 		return err
 	}
+	// Rollback lever for the geographic diversity slots in the ranked relay
+	// page; enabled unless explicitly turned off (and inert in legacy ranking).
+	pageDiversity := envEnabled("OPENRUNG_RELAY_PAGE_DIVERSITY")
 
 	// Fail closed: with no registration token, anyone can register a relay and
 	// poison the directory that clients route their VPN traffic through. Require
@@ -146,6 +149,8 @@ func run() error {
 		GeoIP:                      geoResolver,
 		SigningSeed:                signingSeed,
 		WSSTicketSigningSeed:       wssTicketSeed,
+		RelayRanking:               rankingMode,
+		RelayPageDiversityDisabled: !pageDiversity,
 	}
 	// The Postgres store aggregates dashboard queries in SQL; the JSONL sink's
 	// dashboard is aggregated in Go from its in-memory record set.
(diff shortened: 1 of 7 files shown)


openrung-admin:
Pushed 3b561aa addressing the four review findings:

1. **WSS reservation clobbering the diversity fill** — both passes wrote slot `limit-1`, so on the live fleet shape (WSS-less volunteer head, WSS-less EU fill, WSS-capable foundation relay below the fold) `reserveWSSCandidate` overwrote the fill and the feature was a byte-identical no-op for exactly the Iran→EU corridor it exists to fix. `diversifyRelayPage` now detects that the reservation will claim the last slot (head has no WSS-eligible relay, no fill provides one, one exists below the fold) and starts its fills at `limit-2`. Pinned by `TestDiversityFillSurvivesWSSReservation`, which composes both passes on that fleet shape and requires the EU fill *and* the WSS relay on the final page.
2. **Sole-representative eviction** — victim selection now tracks per-region counts on the page and skips any slot holding a region's last representative, so a fill can never trade one missing region for another; coverage only grows. If every displaceable slot is protected, the fill is dropped and the page stays the global head.
3. **Region map coverage** — the map now spans all the fleet's deploy-target countries (added US/BR → americas, AU → oceania, HK/ID → asia, per `deploy/relay/aws/wss-origin-targets.json`), and `resolveRelayGeo` logs a warning whenever a relay's country resolves outside the map, so the map falling behind the fleet is visible instead of silently shrinking coverage.
4. **Fail-open rollback flag** — `OPENRUNG_RELAY_PAGE_DIVERSITY` now parses fail-closed: empty/truthy enables, the full falsy vocabulary sibling knobs accept (`0/false/no/off/disabled/none`) disables, and anything else refuses to start — a mid-incident typo crash-loops visibly instead of silently leaving the feature running. Documented in `deploy/broker/.env.example`.

Full repo test suite green.

PR #146, opened 2026-08-21
Restore public-IP geo telemetry wiring in the connectcore engine

openrung-admin:
## Summary

`connectcore/clienttelemetry/geoip.go` (`LookupGeoAttributes`) and `Manager.SetGeoAttributes` were unreachable from all production code. Git history shows why: the pre-ADR-001 `cmd/client/main.go` called

```go
go mgr.SetGeoAttributes(clienttelemetry.LookupGeoAttributes(ctx, nil))
```

right after `BeginSession`, but the B1 TUI rewrite (8227a3e, PR #132) deleted that flow and the engine never picked it up. CLI telemetry sessions have shipped without country/city/ISP dashboard attribution since ~2026-08-16 — a silent regression. (Desktop never had the wiring; mobile has its own GeoIpClient and is unaffected. PR #145's module promotion just moved the already-dead code.)

## Change

- `connectFlow` now spawns `attachGeoAttributes(ctx, mgr)` right after `BeginSession`/`connection_attempted` — concurrent so connect never waits on it, and **before the tunnel/OS proxy comes up** so the lookup reports the client's geo, not the relay's (constraint documented on the helper).
- `lookupGeoAttributes` is a package var (mirroring `var clientID`) so tests can stub the ipwho.is lookup.
- Tests: stubbed lookup → real `Manager` → flush through a capturing transport, asserting flushed events carry `country`/`country_code`/`isp`; plus a nil-manager guard test proving no HTTP lookup fires without telemetry.
- `connectcore/VERSION` 0.1.0 → 0.1.1 (nested module; CI enforces the bump). Merging auto-tags `connectcore/v0.1.1` — the mobile re-vendor PR #98 pin may want the new tag.

## Behavior note

Wiring at the engine level means **desktop and every other connectcore host now send geo attributes for the first time** (parity with mobile), not just the restored CLI. If desktop should stay geo-less, say so and I'll gate it.

## Test plan

- `go vet ./...` and `go test ./... -count=1` in `connectcore/` pass (all packages).
- Root module still builds.


openrung-admin:
Addressed the review in a4cc4ba — both timing holes are closed with explicit joins rather than scheduling:

- `attachGeoAttributes` now returns a done channel and runs the lookup on its own bounded context (5s backstop over the 4s HTTP timeout), decoupled from the connect context so a cancelled/failed connect can still stamp its terminal events.
- **Pre-tunnel barrier**: `connectFlow` joins the lookup *before `runLadder`*, not before `promote` — in TUN mode the ladder installs routes as each candidate is tried (readiness waits for them), so pre-promote would already be too late. This is the last point before any candidate can capture traffic; the broker fetch + latency ranking run concurrently with the lookup first, so the join is usually free and always bounded.
- **Terminal-flush barrier**: `finalizeConn` joins `conn.geoDone` before the terminal records (`connection_failed`/`tunnel_stopped`) and flush, so a fast ladder failure or early disconnect no longer loses geo. The tunnel is already torn down at that point, so a late-completing lookup still observes the client's own network.

New timing tests: the done channel must not close while the lookup is in flight and must guarantee attachment once joined; and `finalizeConn` flushes `connection_failed` carrying geo when the lookup completes only after the failure (lookup gated to release while finalize is already waiting). Full `connectcore` suite passes under `-race`.

openrung-admin:
Round 2 addressed in the latest commit — nothing waits on the lookup anywhere anymore:

- `attachGeoAttributes` returns a `geoLookup` handle; **`abandon()` cancels, never waits**, and is nil-safe/idempotent.
- **Ladder boundary**: the pre-`runLadder` join is replaced with `conn.geo.abandon()`. Correctness is preserved without blocking: a result that already arrived was by definition fetched pre-tunnel and is kept; only in-flight I/O is cancelled, so nothing fetched through the tunnel can ever be recorded. A slow/blocked ipwho.is now costs zero connect latency — the lookup just loses the race and the session goes geo-less (matching the old cmd/client's best-effort behavior).
- **finalizeConn**: the unconditional wait is gone; it abandons instead, so teardown/UI and the terminal flush proceed immediately and no lookup outlives its session. A session that ends before the lookup completes flushes without geo — accepted as the best-effort trade.
- On the server-side alternative (broker resolving geo from the observed pre-tunnel IP): agreed it's the more robust design, but it's a broker+ingest change and ADR-002 already constrains how client IPs may be geolocated — left out of this PR deliberately.

Tests: abandon releases a lookup blocked until cancellation (well under the 5s backstop) and stamps nothing afterwards; `finalizeConn` returns immediately with a stuck lookup, flushes `connection_failed`, and cancels it. Root + connectcore suites pass under `-race`.

PR #151, opened 2026-08-23
Client TUI: i18n (EN/中文/русский), country flags, solid status bar, and quoted sing-box crash errors

openrung-admin:
Started as internationalization and grew into a broader TUI overhaul plus one engine-level diagnostics fix, one commit each.

## 1. Internationalization — English, Chinese, Russian

- **No settings menu**: the `5` key cycles the language in place (en → 中文 → русский → en) without leaving the current view.
- The header's fifth slot reads **`5 languages/中文/языки`** — deliberately identical in every language, so a user who cannot read the active language can always recognize the switch.
- All view chrome routes through a per-language `translation` table (`cmd/client/i18n.go`). **English entries are byte-identical to the previous hardcoded copy**, so default rendering is unchanged and the pre-existing TUI tests pass untouched.
- Deliberately still English: engine log lines, error strings, and notice reasons (they quote sing-box output and flags), plus the platform TUN parentheticals shared with `--tun` flag help.
- A reflection test guards the tables: fails on any string added to one language but left empty in another, or on fmt-verb-count drift (which would render `%!s(MISSING)` only for non-English users). Chinese uses indexed verbs (`%[2]s`) where word order swaps arguments.

## 2. Country flags + label-only relay list

- Country codes carry their flag emoji (`JP🇯🇵`) in the Status view and the Relays COUNTRY column.
- Relay-list cells pad by display width (`padCell` over `lipgloss.Width`) instead of `%-8s`, so double-width emoji and CJK headers can't break column alignment.
- The Relays list names a relay by its friendly label alone (`merry-falcon`) — the COUNTRY column already carries the geography. The `City, Country` geo name remains only as the fallback for unlabeled relays.

## 3. Solid tmux-style status bar

- The footer is a full-terminal-width colored bar (black text on background): **red** disconnected/failed, **yellow** through transitions, **green** connected.
- While connected, the bar's right corner shows `<relay label> <flag> HH:MM:SS`, live off the existing 200ms tick.
- On terminals too narrow for both, the help text truncates (width-aware, CJK-safe) and the connection summary wins.

## 4. Quote sing-box's own error on abnormal exit (`connectcore/client`)

Found via a real failure: a `go build` without `-tags with_utls` makes every connect die with the opaque `sing-box exited: exit status 1`, the actual reason buried in the Logs view. The runner now tees each stream through a `crashLineRecorder` (streams recorded separately so interleaved chunks can't splice half-lines; last error/fatal/panic-looking line preferred, last non-empty line — newline-terminated or not — as fallback; 512-byte line cap; shim's `error: ` prefix dropped) and quotes it:

```
local VPN setup failed: sing-box exited: create sing-box service: initialize outbound[0]:
uTLS, which is required by reality is not included in this build, rebuild with -tags with_utls (exit status 1)
```

Verified end-to-end with an untagged build. Desktop inherits the improvement through the shared runner. `connectcore/VERSION` 0.1.1 → 0.1.2. A follow-up commit fixed a test-only data race (`-race` on CI): the crash test pointed exec's two copy goroutines at one bare `strings.Builder`.

## Tests

New coverage: language cycling and header label invariance; translation-table completeness/format parity; flag mapping incl. invalid codes; label-only listing with geo fallback; connected/disconnected footer content; narrow-terminal precedence; footer style coverage of every status; crash-line quoting with the log tee intact; unterminated-last-line fallback. Full `cmd/client` and connectcore module suites pass under `-race`; `go vet` and `gofmt` clean.

Known limitation: language is not persisted across launches (would need a `clientstate` field) — happy to follow up if wanted.


openrung-admin:
Second commit (`d25db2c`) extends this PR beyond i18n with three UI upgrades:

1. **Country flag emoji** next to country codes (`JP🇯🇵`) in the Status view and the Relays COUNTRY column. Relay-list cells now pad by display width (`padCell`) instead of `%-8s`, so the emoji can't break column alignment.
2. **Relays list shows the friendly label only** (`merry-falcon`), not `Tokyo, Japan` — the COUNTRY column already carries the geography. The geo name remains solely as the fallback for unlabeled relays.
3. **Status-colored footer**: red disconnected/failed, yellow through transitions, green connected. While connected the bottom-right corner shows `<label> <flag> HH:MM:SS`, live via the existing 200ms tick. On terminals too narrow for both, the help text truncates (width-aware, CJK-safe) and the summary wins.

New tests cover the flag mapping, label-only listing with geo fallback, the connected/disconnected footer, narrow-terminal behavior, and that `footerStyles` covers every status.

openrung-admin:
Third commit (`067abc5`) reaches outside the TUI into `connectcore/client`: when sing-box exits abnormally, the runner now quotes the child's own last error line in the returned error instead of a bare exit status.

Before (Status view Error row):
```
local VPN setup failed: sing-box exited: exit status 1
```
After:
```
local VPN setup failed: sing-box exited: create sing-box service: initialize outbound[0]: uTLS, which is required by reality is not included in this build, rebuild with -tags with_utls (exit status 1)
```

Mechanics: each stream (stdout/stderr separately, so interleaved chunks can't splice half-lines) is teed through a `crashLineRecorder` that remembers the last error/fatal/panic-looking line, falling back to the last non-empty line — including an unterminated final line, since crashing processes rarely end with a newline. stderr is preferred (both sing-box and the bundled run shim report there), the shim's `error: ` prefix is dropped to avoid stuttering inside the error chain, and line memory is bounded at 512 bytes. The tee still delivers everything to the engine log unchanged.

Found while diagnosing a real user-facing failure: a `go build` without `-tags with_utls` produces a client whose every connect dies with the opaque exit-status-1, with the actual reason buried in the Logs view. Verified end-to-end with an untagged build — the Error row now carries the full rebuild instruction. Desktop inherits the improvement through the shared runner.

Tests: crash quotes the child's words while the log tee still receives all output; no-error-line crash falls back to the unterminated last line. `connectcore/VERSION` 0.1.1 → 0.1.2.

openrung-admin:
`352cdd5` addresses all four review findings:

1. **firstCrashLine stream priority** — now two-tier: every stream is checked for an error-looking line first (stderr still wins ties), and only if none exists does any stream's last non-empty line get quoted. New test: stderr chatter + stdout `FATAL:` → the FATAL line is quoted.
2. **Footer wrapping while disconnected** — the help now truncates against the terminal width unconditionally, not just when a connection summary exists, and the composed line is hard-capped for terminals narrower than the summary itself.
3. **Header wrapping** — trailing tabs are dropped whole once the bar exceeds the terminal width (the number keys still work); headerHeight's one-line budget holds.
4. **Rune-count relay-name truncation** — names (and the recents line) truncate by display cells via truncateWidth; a 28-rune CJK label can no longer push the COUNTRY/LATENCY/CLASS columns out of alignment. The rune-count `truncate` helper is deleted. New test renders a CJK-labeled row next to an ASCII one and asserts equal row widths.

Also new: a sweep test asserting header and footer stay within 20/40/60-column terminals across all three languages and both connection states. Both suites pass under `-race`.

openrung-admin:
`bdafcdf` addresses both findings — shedding now goes by importance, not position:

1. **Header** keeps the active tab and the language control over the title, and the title over inactive view tabs; whatever fits renders in bar order. Rendered on Settings:
   - 60 cols: ` OpenRung  1 Status  4 Settings  5 languages/中文/языки`
   - 40 cols: ` 4 Settings  5 languages/中文/языки`
   - 20 cols: ` 4 Settings`
2. **Footer summary** fits itself to a budget with the duration anchored at the right edge and the relay label giving way first; 20 cols renders `merry-fa… 01:23:00`.

The narrow-width sweep test now also asserts the priorities: active tab visible at 20/40/60 in all three languages, language control at ≥40, and the full session duration at every tested width while connected.

openrung-admin:
Ran an adversarial self-review over this PR's diff (8 finder angles + verification); `1b17984` fixes the top six findings:

1. **Crash quote kept out of broker telemetry** — the runner now returns a typed `crashError`; `Error()` carries the child's words for the TUI/desktop, while a `TelemetrySafe()` rendering carries only `sing-box exited: exit status N`. `clienttelemetry.ErrorDetail` walks the chain and substitutes the safe rendering (wrapping context preserved), so `failure_detail` can no longer upload local paths/usernames a crash line might quote.
2. **os/exec same-writer contract restored** — when a caller passes one writer for both streams, the runner wraps it in a single shared recorder, keeping exec's serialized-Write guarantee for plain unsynchronized writers (module consumers were exposed to a data race).
3. **Stale-error misquoting fixed** — quotes come from a ring of the child's final 8 lines, so a recovered transient `ERROR` from long before the exit can't be blamed for the crash.
4. **UTF-8-safe long-line flush** — the 512-byte forced flush backs off to a rune boundary; no more potential mojibake in quoted CJK-bearing errors.
5. **Footer label/flag single-sourced** — both now come from the polled connection descriptor, so a failover can't render the new relay's name beside the old relay's flag.
6. **Language-stable notices** — settings notes and the shell-unavailable notice are stored as kinds and worded at render time through the active language, so they follow a `5`-key cycle instead of leaving mixed-language UI.

Remaining review findings, deliberately not fixed here: the `…` East-Asian-ambiguous-width cosmetic, the width≤2 summary-budget edge (masked by Bubble Tea's own line truncation), an explicit Error-row truncation, and language persistence (stated follow-up).

New tests: telemetry-safe rendering (both the crashError contract and ErrorDetail's substitution), same-writer serialization under `-race`, stale-error ring behavior, rune-boundary flush, and notes following language cycles. Both modules' full suites pass under `-race`.

openrung-admin:
`9722f7a` addresses both findings:

1. **Footer identity strictly single-sourced** — with a descriptor (`infoOK`), every field is the descriptor's: name via `relayListName` down to its `relay <id>` fallback, plus flag. The state label appears only when no descriptor exists, and then flagless — "Singapore 🇯🇵" is impossible. Regression test covers both the nameless-stale-descriptor case (renders `relay r9 🇯🇵`, never the state label) and the descriptor-less case (bare state label, no flag).
2. **Per-stream line assembly in shared-writer mode** — each stream always gets its own recorder, so a partial stderr line interleaved with stdout output assembles whole instead of splicing into cross-stream mush; the caller's single writer is serialized through an explicit mutex wrapper (`syncWriter`), replacing the reliance on os/exec's same-writer path. New test interleaves a split stderr line with stdout chatter and asserts the quote is `real stderr failure continues`, chatter-free — alongside the existing `-race` shared-writer test.

PR #160, opened 2026-08-24
Windows TUN mode: stdin-close stop channel for the bundled sing-box runtime

openrung-admin:
Lifts the Windows TUN refusal (`cmd/client/elevation_windows.go`) by giving the engine a graceful-stop channel that needs no signal delivery: the bundled runtime now stops when its stdin closes.

## Why stdin, not console events

The old gate existed because a TUN child must be stopped gracefully — only sing-box's own stop handling removes the routes and DNS it installed — and Windows can deliver no stop signal (`os.Interrupt` is unsupported; no console control event reaches a `CREATE_NO_WINDOW` child). But since #149 the "sing-box binary" is our own executable re-exec'd, so we control both sides of the boundary and don't need Win32 console machinery at all:

- **Runner** (`connectcore/client`): when `StopOnStdinClose` is set, appends `-stop-on-stdin-close` to the child argv, holds a stdin pipe open for the child's lifetime, and closes it on context cancel. The interrupt-then-hard-kill ladder is unchanged as the backstop.
- **Shim** (`internal/singboxruntime`): with the flag (the constant is imported from the runner package, so the two sides cannot drift), stdin EOF cancels the same context the signals cancel → `instance.Close()` unwinds routes/DNS → exit 0. Flag opt-in keeps a standalone `run -c cfg </dev/null` from exiting at launch.

This also fixes a latent gap on **every** platform: the OS closes the pipe when the parent dies for any reason, so a tunnel child orphaned by a crashed/force-killed client now unwinds its own routes and DNS instead of black-holing the machine. Signals never covered that case.

## Commits

1. **connectcore: stdin-close stop protocol in the runner** — `StopOnStdinCloseFlag`, `SingBoxRunner.StopOnStdinClose`, `Engine.SingBoxStopsOnStdinClose`; comment updates on `LadderKillGrace`/`TUNKillGrace`.
2. **client+desktop: shim handler + host wiring** — the stdin watcher in the shim; both hosts declare the protocol when self-exec'ing (never for an external `-sing-box`/`OPENRUNG_SING_BOX` binary, which would refuse the unknown flag); on Windows, `--tun` + external binary is refused (`tunRequiresBundledRuntime`, goos-parameterized, enforced at flag parse and in the Elevation hook for the Settings-toggle path).
3. **connectcore: Windows kill-on-close job object** — orphan backstop (the Unix `Setpgid` counterpart), **scoped after review to children that speak no stop protocol** (external binaries): a kill-on-close job holding the bundled child would terminate it immediately when the dying parent's handles close, racing and defeating the pipe's graceful teardown. Best-effort by design.
4. **client: lift the gate** — `tunModeAvailable()` on Windows now checks for an elevated token (wintun adapter creation needs Administrator) instead of refusing outright; Settings copy becomes "needs Administrator"; docs + notices updated.

## No driver distribution needed

sagernet/sing-tun `go:embed`s `wintun.dll` and loads it **from memory** (`internal/wintun/memmod`) — no sidecar DLL, no extraction, single-file client preserved. THIRD_PARTY_NOTICES §8 already carried the wintun license; its "never loaded" note is updated to "loaded only for Windows TUN mode".

## Tested here (macOS)

- Full root-module + connectcore suites with `-race` and release tags; `GOOS=windows` cross-build **and vet** of `cmd/client` and the desktop app (CI-parity commands from go-checks).
- New end-to-end test drives the **real bundled runtime**: a re-exec'd `run` child brings its mixed inbound up, stdin closes, child exits 0 promptly (`TestRunSubcommandStopsOnStdinClose`).
- Runner-level proof that the EOF (not the signal, not the kill) is what stops a `StopOnStdinClose` child, and that the flag never leaks to an external binary.

## Windows verification checklist (needs a real Windows box)

Build: `go build -tags with_utls,with_external_windivert ./cmd/client` (or a client-release dispatch artifact).

- [ ] Non-elevated terminal: `openrung-client connect --tun` refuses with the "Run as administrator" guidance; Settings Mode row shows "needs Administrator".
- [ ] Elevated terminal: `connect --tun` reaches CONNECTED (wintun adapter appears; `route print` shows the tunnel routes; browsing goes through the relay).
- [ ] `d` disconnect: routes/DNS restored, adapter gone, networking intact.
- [ ] `q` quit while connected: same.
- [ ] Failed-rung teardown: pin a dead relay (`-relay-id`), let a TUN candidate fail; no route/DNS residue afterwards.
- [ ] Force-kill the client (Task Manager / `Stop-Process`) while TUN-connected: the sing-box child should exit **on its own within ~a second, gracefully** (stop-pipe EOF — no job object touches the bundled child) and unwind routes/DNS; networking intact. This is the empirical answer to the old gate's residue fear.
- [ ] Run the real parent-death regression test on the Windows box: `cd connectcore && go test ./client/ -run TestParentDeathLetsTheStopProtocolChildExitGracefully -v` — it terminates a supervisor running the actual runner and fails if anything kills the child before its stdin EOF arrives.
- [ ] `--tun -sing-box <external sing-box.exe>` is refused with the bundled-runtime message.
- [ ] Proxy mode unchanged (connect, disconnect, system proxy restored).
- [ ] Connected footer repaints in place — no stale green bars stacking, and no flag emoji anywhere in the Windows TUI (a visible flag means a stale build).

Windows smoke-run finding and its fix: the connected footer stacked stale green bars — Windows terminals draw a flag-emoji pair wider than width libraries count (Segoe UI Emoji has no flag glyphs), overflowing the width-padded footer so it wrapped and scrolled every repaint. The client now never renders flag emoji on Windows (country codes remain). Checklist item below verifies it. The English-only TUN refusal text is the documented i18n policy (platform TUN copy and engine errors stay English on every platform — i18n.go).

Expected noise while TUN-connected: repeated `router: UDP is not supported by outbound: proxy` errors — captured QUIC/UDP being rejected because the VLESS outbound does not carry UDP (pre-existing on every TUN platform; browsers fall back to TCP). Not a Windows failure.

Notes: no VERSION bump here (release-time bump PRs stay separate, as with #156); connectcore change is additive and this PR bumps connectcore/VERSION to 0.1.3 (merge = tag, per go-checks policy); mobile does not consume `SingBoxRunner`, so no mobile re-vendor is needed for this.


openrung-admin:
Review P1 addressed in 3753002: confirmed — the kill-on-close job raced the stop pipe's graceful orphan teardown (handle-close order on parent death is undefined, and closing the job's last handle terminates the child immediately, per Microsoft's docs). The job object is now applied only to children that speak no stop protocol (external binaries, proxy-only on Windows); the bundled child's parent-death cleanup is the stop pipe alone. Accepted tradeoff, documented in the code: an orphaned bundled child that also wedges and ignores EOF is not reaped. The requested real-Windows regression test is in `connectcore/client/engine_windows_test.go` — it terminates a supervisor running the actual runner (TerminateProcess, as Stop-Process does) and fails unless the child exits gracefully on its own; it runs in the Windows verification pass and is compile-checked by the GOOS=windows vet in CI.

openrung-admin:
Review P2 addressed in 9006cfc — confirmed real, and it also invalidated a claim in my earlier P1 response: I said the Windows test was "compile-checked by the GOOS=windows vet in CI", but that vet was run locally; CI's Windows steps only `go build` cmd/client, which compiles no test files. Nothing in CI compiled or executed engine_windows_test.go. go-checks now has a targeted `go-windows` job on windows-latest running `go vet ./client/ && go test -v ./client/` from connectcore, so the parent-death teardown assertion (and every other Windows-only engine path) executes on every PR. The job uses a bash shell deliberately — pwsh run blocks surface only the last command's exit code, which would let a vet failure slip past a passing test.

PR #163, opened 2026-08-25
TUI: old-terminal green-on-black theme, corner brand mark, and slimmer chrome

openrung-admin:
Reskins the terminal client as an old-terminal display and trims its chrome down to what a TUI actually needs. Four commits, all confined to `cmd/client`.

## Green-on-black theme

`paintFrame` lays an old-terminal base — green text on a black background — under the composed frame:

- Every line pads to the full terminal width, so the black background reaches the right edge on all rows (including blank separators). The frame is exactly `m.height` lines, so the whole alt screen is painted.
- Every reset an inner style emits is re-opened with the base sequence, so text after a styled segment falls back to green-on-black instead of the terminal's own defaults. Verified against real frames: lipgloss v1.1.0 emits only `\x1b[0m`, and every occurrence in all four views is followed by the base.
- Styles that deliberately pick their colors are untouched: the solid status bar keeps its red/yellow/green connection signal, and errors, notes, and badges keep theirs.
- Skipped on the Ascii color profile (dumb terminals, `NO_COLOR` — and tests, which keep comparing unstyled output).

## Corner brand mark

An ASCII-art **OPENRUNG** (figlet small, 42×4) is drawn at the bottom right of every view, just above the status bar:

```
                                                 ___  ___  ___ _  _ ___  _   _ _  _  ___
                                                / _ \| _ \| __| \| | _ \| | | | \| |/ __|
                                               | (_) |  _/| _|| .` |   /| |_| | .` | (_ |
                                                \___/|_|  |___|_|\_|_|_\ \___/|_|\_|\___|
```

It renders whole or not at all — hidden on terminals too narrow or short to spare the corner — and the cut underneath is ANSI-aware (`ansi.Truncate`), which closes any open style at the cut so the mark cannot inherit the body's color.

**Known limitation, see Follow-ups:** the mark currently *overlays* the body, so it occludes content rather than yielding to it.

## Language switch: `.`, and out of the tab bar

The tab bar now holds only views. The language cycles in place on a **period** — `5` and the header's `5 languages/中文/языки` slot are both gone — and the footer advertises it with the same trilingual token in every translation, `. lang/语言/язык`.

Punctuation rather than a letter, because this key must be *typeable* in the same situation it must be *readable*. Cyrillic (ЙЦУКЕН), Greek, Hebrew, and Arabic layouts carry no Latin letters at all, so a letter binding is unreachable without first switching the OS layout — exactly what a user stuck in a language they cannot read is least able to work out. A period sits on essentially every layout (unshifted on ЙЦУКЕН and QWERTZ, shifted on AZERTY, all delivering `.`), the same property that made the old `5` safe. `。` is bound alongside it for a CJK IME in full-width punctuation mode.

Because broker URLs are mostly periods, there is a test pinning that a focused settings editor swallows the key as text instead of cycling the UI out from under the value being typed.

## Chrome trimmed

- Footer help drops `1-4/tab views` — the tab bar already shows them.
- The Relays view drops the recents line. The engine still persists recents (desktop parity); the TUI just stops rendering them.
- Settings drops the target relay id / label / country editors, and is down to broker, mode, and the shell helper. Targeting stays where it is useful: `enter` on a relay in the Relays view pins it, `x` clears, CLI flags still apply, and the Status view's Target row shows what is pinned.

## Tests

Brand mark placement and right-alignment on all four views; whole-or-nothing hiding on small terminals; the ANSI-aware cut against full-width styled content; `paintFrame` padding, base-wrapping, and reset re-opening; the period cycling all three languages (and `。` doing the same, and `5` no longer doing anything); the period staying text inside an editor; and Settings offering no target fields.

`go build ./... && go test ./...` green across the module (`connectcore` untouched). `go mod tidy` promoted `charmbracelet/x/ansi` and `muesli/termenv` from indirect to direct — no new dependencies.

## Follow-ups (reviewed, not yet fixed)

A review of this branch turned up two issues worth fixing before or soon after merge. Both are real; neither is addressed here.

1. **The brand mark occludes load-bearing content.** It overwrites a 43×4 region of the body, so at 80 columns the last four relay rows lose their LATENCY, CLASS and `▸ target` columns — including on the *selected* row when the cursor sits at the tail — and the four newest log lines are cut at 36 cells. Because the log viewport auto-follows to the bottom, the newest lines are always under the mark. This truncates the sing-box crash line that #151 added quoting for. Measured widths that would clear the mark: 116 columns for Relays, 126 for Settings; the Logs view cannot be fixed by width at all, since its lines are unbounded. The durable fix is to *reserve* the band — subtract its rows from `bodyHeight` and draw it in its own strip — so the mark costs 4 rows of content instead of 43×4 cells of content.

2. **Footer truncation can hide the language token.** Per-view help is prepended to the global help, so the trilingual token is the first thing lost to truncation: at 80 columns while connected in Russian it is absent from all four views, and even at 120 it is absent from Relays. `TestFooterAlwaysShowsTheTrilingualLanguageHelp` passes only because it exercises the status view, disconnected — the one configuration where the token survives.

Lower priority, from the same review: `describeTarget`'s Country branch and `tr.targetCountry` in three languages are now unreachable (no production writer of `RelayTarget.Country` remains); `paintFrame` and `overlayBrandMark` each re-open-code the existing `padCell`; `settingsFieldValue`/`applySettingsField` are now single-field wrappers; `runTUI`'s comments still justify state seeding by recents that nothing renders; and the pre-existing `truncateWidth` is exactly `ansi.Truncate(s, w, "…")`, which this PR makes a direct dependency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,16 @@ openrung-telemetry.jsonl
 # Claude Code local worktrees (may contain absolute paths / local branch content)
 .claude/agents/
 .claude/worktrees/
-/relay
+
+# `go build ./cmd/<name>` drops the binary in the repo root under the package
+# name. Ignore every cmd/ package's binary: a Go binary embeds the builder's
+# GOMODCACHE paths (/Users/<name>/go/pkg/mod/...), so committing one leaks the
+# local username into public history.
 /broker
+/client
+/frontcheck
+/relay
+/relayhub
+/wssmatrix
+/wsssidecar
 /openrung-*
--- a/cmd/client/headless.go
+++ b/cmd/client/headless.go
@@ -31,12 +31,13 @@ const (
 )
 
 type commonConfig struct {
-	BrokerURL  string
-	Limit      int
-	MTU        int
-	Family     string
-	RelayID    string
-	RelayLabel string
+	BrokerURL    string
+	Limit        int
+	MTU          int
+	Family       string
+	RelayID      string
+	RelayLabel   string
+	RelayCountry string
 }
 
 func parseCommonFlags(name string, args []string) (commonConfig, error) {
@@ -56,10 +57,14 @@ func addCommonFlags(fs *flag.FlagSet, cfg *commonConfig) {
 	fs.StringVar(&cfg.Family, "relay-family", defaultRelayFamily, "relay address family: auto, ipv4, or ipv6 (check/config)")
 	fs.StringVar(&cfg.RelayID, "relay-id", "", "connect only to the relay with this exact broker relay id (e.g. relay_abc...)")
 	fs.StringVar(&cfg.RelayLabel, "relay-label", "", "connect only to the relay(s) with this label")
+	// A country target keeps the engine's mid-session failover WITHIN the
+	// country (an exact relay id has nothing to fail over to) — for networks
+	// where only one country's relays are usable at all.
+	fs.StringVar(&cfg.RelayCountry, "relay-country", "", "connect only to relays in this 2-letter country code (e.g. kr); failover stays within it")
 }
 
 func (cfg commonConfig) target() connectcore.RelayTarget {
-	return connectcore.RelayTarget{RelayID: cfg.RelayID, Label: cfg.RelayLabel}
+	return connectcore.RelayTarget{RelayID: cfg.RelayID, Label: cfg.RelayLabel, Country: cfg.RelayCountry}
 }
 
 func runCheck(args []string) error {
--- a/cmd/client/i18n.go
+++ b/cmd/client/i18n.go
@@ -9,7 +9,7 @@ import (
 )
 
 // The TUI ships English, Chinese, and Russian. There is no settings entry for
-// this: the . key cycles the language directly, and the footer help token that
+// this: the 0 key cycles the language directly, and the footer help token that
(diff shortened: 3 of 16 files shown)


openrung-admin:
Added a second commit: the language cycle moves from `5` to `l`, and the trilingual language slot leaves the tab bar (views only now). The switch is advertised by the footer instead — the same trilingual token `l lang/语言/язык` in every translation, placed at the *front* of the global help because the footer sheds from the right: on a terminal too narrow for the full help (Russian at 80 cols), the language escape hatch is the last thing to go. `5` is fully unbound, tests updated.

openrung-admin:
Third commit, per follow-up feedback:
- Footer help: `1-4/tab views` removed (obvious from the tab bar), the trilingual `l lang/语言/язык` token now sits in its slot. The shorter help fits at 80 cols in every language, so the front-of-help truncation trick from the previous commit is no longer needed.
- Relays view: recents row removed (engine still persists recents for desktop parity; TUI just stops rendering them).
- Settings: target relay id/label/country editors removed — targeting happens from the Relays view (enter pins, x clears) or CLI flags; the Status view's Target row still shows what's pinned. Settings is down to broker, mode, shell helper.

PR #164, opened 2026-08-26
Fix root-owned client state after sudo TUN runs

openrung-admin:
## Problem

`sudo ./client --tun` makes the connectcore engine create `os.UserConfigDir()/openrung` (macOS: `~/Library/Application Support/openrung`, Linux: XDG equivalent) as **root, mode 700** — `os.UserConfigDir` resolves to the invoking user's home while the euid is root. Every subsequent non-sudo run then silently fails to persist recents, the stable proxy-port file (+ its flock file), the telemetry client-id, and the shell-proxy helper; Settings shows `write proxy shell helper: open ...: permission denied`.

Reproduced 2026-08-27 on macOS: `/Users/<user>/Library/Application Support/openrung` was `drwx------ root staff`.

## Fix

New **`connectcore/sudouser`** package: when the process is root **and** `SUDO_UID`/`SUDO_GID` are set, `MkdirAll` chowns every directory it creates to the invoking user and `Chown` hands individual files back. Both are strict no-ops for non-root processes and for genuinely-root daemons (no `SUDO_UID`), so real root services keep root-owned state.

Wired in at every write site under the config dir:

- `internal/clientstate`: `New()` creates the dir owned by the invoking user and, when elevated, repairs a dir/files left root-owned by earlier sudo runs; `writeFile` chowns the atomic temp file before the rename; the flock file is chowned at creation.
- `connectcore/clienttelemetry`: client-id dir/file creation, plus a repair chown when an existing id is read while elevated (a root-owned 0600 client-id would otherwise force ephemeral IDs on every plain run).

`connectcore` 0.1.4 → 0.1.5 (merge = tag).

## Tests

`sudouser` regression tests stub the euid/chown seams so the elevated path runs without root: chown targets the SUDO uid/gid, only newly created directories are chowned (parents first), and no chown happens for non-root, genuine-root, or malformed `SUDO_UID`. Existing `clientstate` and `clienttelemetry` suites pass.

Manual verification still worth doing on a Mac: `sudo ./client --tun` from a clean state, then a plain run — the dir should be user-owned and Settings free of the helper error.


openrung-admin:
Reviewed `907f154` and pushed `1060080` with the three fixes from that review.

**1. The symlink hole was still open in the lock path.** `withFileLock` kept using plain `os.OpenFile(path, os.O_CREATE|os.O_RDWR, ...)` — the `os.Chmod(path)`→`file.Chmod` and `Chown(path)`→`ChownFile(file)` swaps closed the open-to-chown race but not the following open itself, so the descriptor already pointed at the symlink's target. Verified: a symlink at `proxy-port.lock` got its target fchmod'd 0644→0600, and the `ChownFile` on the next line hands that target to `SUDO_UID`. Now routed through `sudouser.OpenRegularFile` like every other state write. It was masked only because `proxy-port.lock` is in the repair allowlist and `New()` hard-failed first — fixing (3) would have unmasked it.

**2. `O_NOFOLLOW` on every ancestor broke real layouts.** A symlinked `~/.config` (dotfiles checkout) or a home on another volume made every elevated run fail — the tests had to add `canonicalTempDir` to dodge this on macOS `/tmp`. Ancestors are now traversed normally and the safety property moved to where it actually holds: anything handed to the invoking user is opened `O_NOFOLLOW` **and** chowned only when its containing directory already belongs to them. So `~/.config` symlinked at `/etc` still creates `/etc/openrung` but never hands it over — the user gains nothing they didn't already have. Two related gaps closed while in there: `ChownFile` refuses a multiply-linked regular file (a hard link reaches a target `O_NOFOLLOW` can't see, and `fs.protected_hardlinks` isn't universal), and regular-file opens carry `O_NONBLOCK` — confirmed `OpenRegularFile` blocked indefinitely on a FIFO at a state path, hanging startup rather than erroring.

**3. Best-effort repair restored.** `New()` had turned five repairs into `return nil, err`, and both callers do `if store, err := clientstate.New(); err == nil` — so one unrepairable entry silently cost the run its recents, stable proxy port, crash snapshot, and shell helper. That is the exact symptom this PR exists to fix. Only creating the directory is fatal again.

Tests added for the symlinked ancestor, a parent the invoker doesn't own, a symlinked state dir, the hard link, the FIFO, and the lock path; the fixed `501:20` pair in the existing tests now derives from the real uid so the ownership gates are exercised on any host. Both modules green, `go vet` clean, windows/linux cross-builds clean. `connectcore` stays at 0.1.5 (unreleased, no tag yet).

The fd-based chown, the allowlist with its canonical-port check, and the Windows shims from `907f154` are all kept as-is — the approach was right.

openrung-admin:
All three confirmed and fixed in `a44e6d9`. I reproduced each before changing anything.

**1. Unsafe state directory returning a usable Store.** Correct, and my change caused it — making `OpenStateDir` best-effort left a `Store` rooted at a symlinked path, and `NewInDir` + `SaveProxyEnvScript` writes straight through it. Refusing the directory is fatal again; only `RepairOwner`/`RepairEntry` stay best-effort, which keeps a repair hiccup from costing the caller its whole store.

Validating at `New` alone isn't enough though: the attacker owns that directory, so `mv openrung openrung.real && ln -s /etc/profile.d openrung` between `New` and the first write wins the race trivially. So the directory is now pinned by descriptor for the process lifetime (elevated runs only) and writes go through `openat`/`renameat` relative to it — a swap after validation can't redirect them. Regression test does exactly that swap and asserts the file lands in the pinned directory. The unelevated path is untouched, and a test asserts both paths produce identical bytes and mode, since the desktop app and TUI share these files.

**2. Hard links rejected after destructive operations.** Correct, and worse than described — the open *succeeded*. `hasExtraLinks` only ran inside `ChownFile`, so `OpenRegularFile` returned a live descriptor to an already-truncated target (verified: 19 bytes → 0, `err=nil`). `O_TRUNC` is now withheld from the `openat` and applied only after the descriptor is proven regular and singly-linked, and the link check moved into `regularFileFromFD` so no mutating caller ever receives a multiply-linked file. That covers the lock path's `fchmod` too — it can't reach a hard link any more, so no reordering was needed there. A companion test asserts the withheld `O_TRUNC` still truncates on the normal path.

**3. `OpenStateDir` breaking symlinked `~/.config`.** Correct — my earlier fix relaxed `MkdirAll`'s traversal but `OpenStateDir` went through `OpenDir`, which pins its own final component, i.e. `.config`. Verified it failed with `openat …/config: not a directory`. The parent path is now followed and only the `openrung` component is `O_NOFOLLOW`; the parent descriptor is still what the ownership gate is evaluated against, at open time. Test covers a dotfiles-style `~/.config -> dotfiles` layout repairing correctly.

`ChownRegularFileAt` and the standalone `OpenStateDir` are replaced by a `StateDir` handle (`RepairOwner`, `RepairEntry`, `Entries`, `WriteFile`), which is what makes the fatal-vs-best-effort split explicit at the call site instead of encoded in one error return. Windows keeps parity shims; `Active` is false there, so repairs are no-ops and writes keep pathname semantics.

Both modules green, `go vet` clean, windows/linux cross-builds and `GOOS=linux go vet` clean. `connectcore` stays at 0.1.5.

openrung-admin:
Both confirmed and fixed in `56d81c3`.

**1. Snapshot deletion bypassing the pin.** Reproduced — with the directory swapped for a symlink after `New`, `ClearProxySnapshot` returned nil having unlinked the target's `proxy-snapshot.json`. `StateDir.Remove` now unlinks relative to the descriptor. It wraps the errno in an `*os.PathError` rather than swallowing ENOENT, so `ClearProxySnapshot`'s existing `os.IsNotExist` check still treats an already-absent snapshot as success; there's a test for that alongside the escape test.

**2. PID-only temporary names.** Reproduced — 13 to 16 of 16 concurrent writes failed with ENOENT on `.proxy-snapshot.json.tmp-<pid>`. My `os.CreateTemp` replacement dropped the per-call uniqueness that the comment above it explicitly promised for concurrent refreshes. Names now carry a process-wide atomic sequence on top of the pid and are created `O_EXCL`, with a retry so a leftover from a crashed run on a recycled pid doesn't wedge the write. The regression test runs 16 goroutines writing distinct helpers plus the same snapshot, asserts every helper holds its own content, and checks no temporary files survive; it passes under `-race` and repeated `-count`.

One more surface closed while in there, since it was the last one left: the three `Load*` methods still read by pathname, and `os.ReadFile` as root will follow a symlink at a state path and block in `open(2)` on a FIFO left there — a startup hang needing no race at all. They go through the descriptor now. Each caller already treats any error as "absent state", so behavior is unchanged; a test confirms a symlinked `recents.json` reads as empty rather than through the link.

That makes the invariant complete and easy to keep: under sudo, every read, write, and delete on the state directory goes through the descriptor validated in `New`. Nothing in `clientstate` touches those files by pathname any more except the unelevated fallback.

Both modules green including `-race`, `go vet` clean, windows/linux cross-builds and `GOOS=linux go vet` clean. `connectcore` stays at 0.1.5.

PR #168, opened 2026-08-28
ADR-003 A1: extract the TunnelRuntime seam from the connect engine

openrung-admin:
## What

Track A, PR A1 of ADR-003 (mobile connectcore adoption): the engine no longer constructs or signals a sing-box process anywhere — every start/stop/restart path (initial ladder, WSS-fallback rungs, failover re-ladder, disconnect/shutdown teardown) runs through a new `TunnelRuntime` seam owned by connectcore. The subprocess implementation is the default and desktop/TUI behavior is unchanged; Track B's in-process libbox runtime plugs into `Engine.TunnelRuntime`.

## Open decision resolved here: `TunnelRuntime` shape

Implemented **shape 2 — factory plus per-run handle** (the maintainer leaning in the ADR):

```go
type TunnelRuntime interface {
    Run(ctx context.Context, configJSON []byte) (TunnelRun, error)
}
type TunnelRun interface {
    Done() <-chan error            // that run's exit report, exactly once, then closed
    Stop(grace time.Duration) error // graceful-then-forced; blocks until the run ends
}
```

Cross-cutting requirements from the ADR, all honored:
- config is always **bytes**, never a path requirement (the subprocess impl materializes and owns the temp file);
- `Run` returns at *launched*, not *ready* — readiness stays the engine's `awaitTunnelReady`;
- context cancellation aborts a launch; a live run's lifetime belongs to `Stop`;
- launch failures and `Done` exit errors feed the shared classifier unchanged (crash-line quoting and the telemetry-safe rendering preserved).

Every attempt gets a fresh handle, so a `Done` report is never ambiguous across the WSS-fallback and failover restart paths, and the in-process implementation's between-runs teardown obligation is explicit.

## How it lands

- `connectcore/tunnelruntime.go` — the interface pair, the subprocess default (`writeTempConfig` + `client.SingBoxRunner.Start`, file removed after that run's exit), and the engine's runtime resolution.
- `connectcore/client/engine.go` — `SingBoxRunner.Run` split into `Start` + `SingBoxProcess{Done,Stop}`; the supervision specifics (stdin-close stop pipe, interrupt/kill ladder + `singBoxHardKillWait`, crash-line recorders, Windows kill-on-close job object, process-group semantics) moved verbatim into the handle. `Run` is now the same primitives composed, still used by `cmd/wssmatrix`.
- `connectcore/engine.go` / `monitor.go` — `candidateResult` holds `run`/`runDone`/`stopGrace` instead of `runErrCh`/`reaped`/`configPath`; teardown stops the run through the seam with the same grace values as before (`LadderKillGrace`, `TUNKillGrace` in TUN mode) and keeps the pinned ordering (punched path closes only after the run has ended).
- The unexported `runTunnel`/`writeConfig` test seams are deleted; tests fake the tunnel through the exported seam.
- `connectcore/VERSION` 0.1.5 → 0.2.0 (Track A opens 0.2.x per the ADR).

## Acceptance (from the ADR prompt)

- **Desktop + TUI suites green** — see test matrix below; golden configs untouched (builder not modified, golden tests pass).
- **No engine code constructs or signals a process except through the seam** — `SingBoxRunner` appears in the root engine package only inside `tunnelruntime.go`'s default implementation.
- **Stub in-memory `TunnelRuntime` exercises the full ladder without any sing-box binary** — `TestStubTunnelRuntimeDrivesFullLadder`: failed first rung (relay-scoped probe failure), promotion of the second candidate, disconnect teardown; asserts the runtime received the generated config as bytes and exactly one `Stop(LadderKillGrace)` per run.

## Resolved divergences

- **Temp-config lifecycle moved behind the seam.** Before: the engine wrote the file and candidate teardown removed it after reaping. Now the subprocess runtime writes it and removes it once that run's process has exited. Net effect is the same on every normal path; on the pathological unkillable-child path the file now lingers with the child instead of being removed under a possibly-live process. The `config_file` telemetry stage is preserved (the runtime marks it; other failures default to `tunnel_start` as before).
- **The candidate context no longer stops the process.** Cancellation used to reach the runner's ctx; now `teardown()` calls `Stop(grace)` immediately after cancelling the context, with the same grace values — same call sequence, one explicit step instead of a ctx side effect. `Run(ctx, …)` also refuses to launch when ctx is already cancelled (before, a child could spawn and be immediately stopped).
- **Stop-pipe read end on the hard-kill-timeout path** now stays open until the child actually exits (held by the supervision goroutine) instead of being closed when `Run` returned. The write end — the EOF that drives the graceful orphan teardown — closes at the stop request exactly as before.
- The WSS Reality test's temp-file-cleanup assertion moved to the subprocess runtime's own end-to-end test (`TestSubprocessRuntimeMaterializesConfigAndCleansUp`), since the engine test no longer sees a file.

## Tests run (macOS)

- `connectcore`: `go vet ./...`, `go test -race ./...` — all packages pass, including the new seam tests and `GOOS=windows go vet ./client/`.
- Root module: `go vet ./...`, `go vet -tags with_utls,with_external_windivert ./...`, `go test -race -tags with_utls,with_external_windivert ./...` — pass.
- `desktop`: vet + `go test -race ./vpnservice/...` + build; Windows app cross-build with placeholder dist (`-tags desktop,production,with_utls,with_external_windivert`) — pass.
- `desktop-volunteer`: vet + `go test -race` over `directsetup`, `volunteerservice`, `persist` + build — pass.
- Windows client cross-build (`GOOS=windows … ./cmd/client`) — pass.
- Not run locally: the real-Windows client suite (`engine_windows_test.go`) — CI's `go-windows` job covers it; the parent-death test uses only the unchanged `Run` surface.

## Notes

- ADR-003 gates Track A on the 0.3.8 soak (tagged 2026-08-23) plus its telemetry confirmation; this PR was opened at the maintainer's direction — merge timing stays with the gate owner.
- After merge, `connectcore/v0.2.0` tags automatically; the mobile re-pin belongs to Track B (B1), not to this PR.


openrung-admin:
Both review findings addressed in 63cfad4:

1. **Stop now completes the run's cleanup.** `subprocessRun` gains a `cleaned` signal, closed by the forwarder right after `os.Remove(configPath)`; `Stop` blocks on it once the child has terminated, so the temp config (relay UUID included) is off disk before `Stop` returns — the same synchronous guarantee the pre-seam teardown gave. A stop that failed to take effect (hard-kill timeout) still returns its error without waiting; cleanup then happens whenever the child finally exits. The end-to-end test now asserts the file is gone the moment `Stop` returns, not after `Done`.

2. **Every launch stage honors cancellation.** `Run` rechecks the context after materializing the config (removes the file, returns the context error) and after `Start` (stops the just-launched child via `Stop`, which also waits for its cleanup, then returns the context error) — it can no longer hand back a live run for a cancelled launch. Two regression tests pin cancellation to each stage deterministically with a context whose `Err` flips after a fixed number of checks: `TestSubprocessRuntimeCancelDuringMaterializationLaunchesNothing` and `TestSubprocessRuntimeCancelDuringLaunchStopsTheStartedChild`.

Re-ran: connectcore `go vet` + `go test -race ./...`, root `cmd/client` + `internal/tunnel` with the release tags, and `desktop/vpnservice` — all green.

openrung-admin:
## Review: LGTM (posted as a comment — same account authored the PR, so GitHub refuses a formal approval)

Reviewed the seam design, the `SingBoxRunner` split, every engine call site, and the test migration; re-ran the suites locally at 63cfad4 (macOS): `connectcore` vet + `go test -race ./...`, `GOOS=windows go vet ./client/`, root module build + vet + `go test -race -tags with_utls,with_external_windivert ./...`, and the Windows client cross-build — all pass.

### Verified

- **Seam invariant holds.** `client.SingBoxRunner` appears in the root engine package only inside `tunnelruntime.go`'s subprocess default (the `logWriter` mention is a doc comment). The deleted `runTunnel`/`writeConfig` test seams have no survivors; all ladder/WSS/TUN/notices tests now fake through the exported `TunnelRuntime`.
- **Every error path that owns a run tears it down.** All seven `teardown()` sites check out; teardown's order (cancel → `Stop` → punch close → WSS bridge) preserves the pinned "bridge never closes under a live sing-box" rule, and `Stop` on an already-exited run is a prompt no-op (proven by `TestStartHandleReportsUnrequestedExitAndStopIsSafeAfterIt`).
- **The supervision move is verbatim-faithful.** Stop-pipe lifecycle, interrupt/kill ladder with `singBoxHardKillWait`, crash-line recorders read only after `Wait`, the Windows job-object release (now with a second-chance release on the hard-kill give-up path — an improvement over the old deferred release), and process-group handling all carried over. `Run` is now genuinely `Start` + `Done`/`Stop` composed, and `cmd/wssmatrix`'s use of `Run`+`KillGrace` is untouched.
- **`stopRequested`/`exited`/`done` synchronization is sound** under the mutex + `sync.Once` arrangement; `-race` is clean over the new concurrency tests.
- **Grace budgets pinned per run** (`stopGrace` set before launch, `TUNKillGrace` in TUN mode) match the old `KillGrace` wiring exactly.
- **Credential hygiene:** commit 63cfad4's guarantee that `Stop` returns only after the temp config (relay creds) is off disk is real and tested (`TestSubprocessRuntimeMaterializesConfigAndCleansUp`, which also correctly polls the stub child rather than sleep-then-signal).
- `VERSION` 0.1.5 → 0.2.0 matches the ADR's Track-A versioning; goldens untouched.

### Nits (non-blocking)

1. **Cancel-during-launch pays the 5s default grace.** `subprocessRuntime.Run`'s post-launch cancel check calls `run.Stop(0)`, which `SingBoxProcess.Stop` maps to the 5s default — not the candidate's `LadderKillGrace`/`TUNKillGrace`. On Windows with an external (stopless) binary, a disconnect that races the launch can block teardown ~5s + the hard-kill wait, where the old ctx-driven path used the engine's short ladder grace. Rare race, correctness unaffected; if it ever matters, thread the grace into the seam's launch-abort path.
2. **Crash racing a Stop reports nil.** `supervise` reads `stopRequested` after `Wait`, so a child that crashes just as `Stop` is requested delivers `nil` on `Done` — the crash line is swallowed. By then the engine has stopped caring about that run, so this is fine; just noting it's the documented semantics ("nil for an exit Stop requested") applied to a genuine crash.
3. On the hard-kill give-up path `subprocessRun.Stop` returns the error without waiting for `cleaned`, so the temp config lingers with the unkillable child — already called out in the PR body and strictly no worse than the old behavior.

Ready for the Codex pass; merge timing stays with the 0.3.8 soak gate owner per ADR-003.

PR #169, opened 2026-08-28
ADR-003 A2: platform hooks — socket control, network signals, explicit lifecycle

openrung-admin:
## What

Track A, PR A2 of ADR-003: the three seams the mobile host needs, each proven on desktop. Every seam defaults to a no-op — desktop and TUI behavior is byte-identical with nothing injected (nil `SocketProtector` collapses every construction to the exact client built before; no `UpdateNetworkState` caller keeps the engine purely probe-driven; `Stop` is now `Shutdown` with the engine's historical defaults).

### 1. Socket control (`Engine.SocketProtector`)

The hook is `wsscore.SocketProtector` — the fd-level, gomobile-friendly shape the shared modules already define (`VpnService.protect`). It now covers every socket the engine opens toward the **physical** network: relay discovery, WSS session tickets, telemetry uploads, the geo lookup, hub punch coordination (`PunchOptions.HTTPClient` added), relay TCP reachability + the ranker's probes, the network-alive gate, and the WSS bridge's CDN dial. It deliberately never touches the through-tunnel internet probe and health sweeps (they verify the tunnel) or loopback dials; the punched QUIC data path protects its own sockets inside the host's `PunchEstablisher`. A refused protection **fails the dial closed** — an unprotected socket would not error, it would silently route into the tunnel.

brokerapi gains `NewHTTPClientWithDialControl` so the protected broker client keeps the ECH / no-SNI dial behavior (the control runs on the TCP socket before any TLS, sharing the authenticated ECH retry-config state) — **brokerapi 0.5.0 → 0.6.0**, with the connectcore/root/desktop/desktop-volunteer requires moved together per the pin procedure.

### 2. Network signals (`Engine.UpdateNetworkState`)

An engine-owned tracker encodes the mobile `PhysicalNetworkEpochMonitor` semantics once, read from the Kotlin and Swift sources as the spec, so the Track B adapters become thin callback forwarders: fingerprint-based change detection (never callback counting), first observation absorbed as the baseline, duplicates ignored, and a per-promoted-candidate re-baseline so an epoch never retires a session it predates (the engine equivalent of Android's fresh per-session monitor).

Consumption: a live **WSS** session is retired by an epoch boundary — orderly, blaming neither the relay nor the front (`transport_session_ended` with a new `trigger=network_epoch` attribute), recovered with the fresh direct-first ladder and a new single-use ticket — exactly the Android invariant ("a WSS socket is bound to one physical-network epoch"). A signal also wakes the recovery network gate immediately instead of waiting out its 5s poll (iOS `waitUntilSatisfied`'s event-driven wait).

### 3. Lifecycle (`Pause` / `Resume` / `Shutdown(flushBudget)`)

`Pause` suspends the engine's periodic self-initiated activity — health sweeps, telemetry heartbeats (the loop moved engine-side over the exported `clienttelemetry.NextHeartbeatDelay`, same cadence), and recovery reactions; triggers that fire while paused are **held, not lost**, and run on `Resume` immediately. Pause never blocks Disconnect/Stop/Shutdown. `Shutdown(flushBudget)` is Stop with a caller-owned bound on the terminal flush and its outcome reported, under a documented order: (1) tunnel stopped through the TunnelRuntime seam + OS proxy restored, (2) terminal state emitted, (3) session-end events flushed within the budget — the only step the budget may cut short (in-memory-outbox drop semantics documented; disk durability is A3's).

## Resolved divergences (vs the mobile sources, recorded per the ADR)

- **Direct/punched sessions on an epoch boundary**: mobile has no epoch monitor on those paths; the engine uses the signal to kick one immediate health sweep instead of retiring a possibly-healthy session — the end-to-end probe stays the authority.
- **Network-down knowledge never vetoes probing**: iOS uses an unsatisfied path to exempt adapter losses from its circuit breaker; the engine's equivalent gate (`networkAlive`) stays dial-probe-driven, so a stale down-signal can accelerate nothing and wedge nothing.
- **Baseline point**: the engine re-baselines at candidate promotion (mobile baselines when starting the per-session monitor right after promote — same semantics, one epoch-counter mechanism).

## Acceptance (from the ADR prompt)

- **Desktop/TUI green** — full matrix below.
- **Pause/resume/network-epoch transitions driven through the seams** — `TestNetworkEpochWhilePausedRebuildsAfterResume` (sleep → epoch → wake → rebuild, the exact mobile sequence), plus `TestPausedEngineHoldsFailoverUntilResume`, `TestPauseHoldsHealthSweepsAndHeartbeatsUntilResume`, and five epoch-semantics tests (baseline/dedup, WSS retirement, direct-path kick, pre-promote epoch, recovery-gate wake).
- **Socket hooks observed end-to-end in a loopback test** — `TestSocketProtectorObservedOnLoopbackConnect`: a real loopback relay dial and a real telemetry upload both reach `Protect` before connecting; `TestSocketProtectorRefusalFailsClosed` proves the fail-closed contract; `TestNewHTTPClientWithDialControlRunsOnEverySocket` covers the brokerapi transport.
- `connectcore/VERSION` 0.2.0 → **0.3.0**; `brokerapi/VERSION` 0.5.0 → **0.6.0**.

## Tests run (macOS)

- `brokerapi`: vet + `go test -race ./...`.
- `connectcore`: vet + `go test -race ./...` (all packages), `GOOS=windows go vet ./...`.
- Root module: vet (plain + `with_utls,with_external_windivert`) + `go test -race` with the release tags.
- `desktop`: vet + `go test -race ./vpnservice/...` + build; Windows app cross-build with placeholder dist.
- `desktop-volunteer`: vet + `go test -race` over directsetup/volunteerservice/persist.
- Windows client cross-build with release tags.
- Not run locally: the real-Windows client suite — CI's `go-windows` job covers it (nothing in this PR touches `connectcore/client`'s process lifecycle).

## Notes

- A3 (absorb the D4 outbox policy + remaining mobile policy richness) builds directly on this: the flush-budget contract here is written against the in-memory outbox and gets its disk durability there.
- The A1 review nit about the launch-abort grace (5s default instead of the ladder grace) is unchanged here — it lives in the subprocess runtime, untouched by this PR; happy to fold a fix into A3.


openrung-admin:
Both findings verified and fixed in 67dc925:

1. **Shutdown now shortens an already-running terminal flush.** Verified: `finalizeConn` captured the budget at flush start, so a Shutdown behind a Disconnect stored a value nobody read and waited out `conn.done` for up to the default 5s. The flush is now a registered, cancellable operation — `terminalFlush` stores its cancel on the connection in the **same lock acquisition** that reads the budget, closing the race in both directions: Shutdown's budget stored first bounds the flush directly; otherwise Shutdown finds `flushCancel` and arms a timer so the in-flight flush gets at most Shutdown's own budget more. Regression test: `TestShutdownShortensAnAlreadyRunningTerminalFlush` — Disconnect parks the terminal flush on a hanging broker (the upload's arrival proves the flush is in flight with its cancel registered), then `Shutdown(100ms)` returns in ~130ms with the interruption reported; before the fix it blocked the full 5s.

2. **Protection refusal is now a local platform failure.** Verified: the ladder dial returned `errSocketProtectionFailed` as an ordinary reachability error → `directPathError` → `relay_attempt_failed` against a relay never actually dialed, plus WSS-fallback unlock. Both the ladder dial (engine sentinel) and the WSS bridge dial (`wsscore.ErrSocketProtectionFailed`) now short-circuit through `localCandidateError` with a `socket_protection` stage — the ladder stops, no relay health is dented, no single-use ticket is minted, and `lastError` still names the protection failure. Regression tests: `TestSocketProtectionRefusalIsLocalNotRelayEvidence` (no `relay_attempt_failed`, no ticket request, local stage in the log) and `TestWSSDialProtectionRefusalStopsLadderWithoutFrontDamage` (one front dial, one ticket, no `transport_failed`, only the genuine direct failure recorded); the loopback fail-closed test now also asserts the local classification.

Re-ran: connectcore `go test -race ./...` + `GOOS=windows go vet`, root `cmd/client`/`internal/tunnel` with release tags, `desktop/vpnservice` — all green.

openrung-admin:
Verified and fixed in the follow-up commit: the ticket path was indeed still front-blaming a protection refusal — `requestWSSSessionTicket` ground through every eligible broker front plus the bounded Retry-After round, and `attemptWSSCandidate` wrapped the outcome as a `ticket`-stage `wssTransportError`, producing one false `transport_failed` per relay front.

The ticket loop now short-circuits on `isSocketProtectionFailure` (one attempt, no cross-front retries, no retry round — every further attempt would reproduce the same host-local refusal), and the caller classifies it `localCandidateError("socket_protection", err)`, identical to the ladder-dial and bridge-dial sites from the previous fix. `TestWSSTicketProtectionRefusalStopsAfterOneAttempt` pins it: two-front relay, refused ticket fetch → exactly one ticket attempt, zero WSS dials, zero `transport_failed`, only the genuine direct failure's `relay_attempt_failed`, and the `socket_protection` local stage in the log.

Re-ran the full connectcore race suite, `GOOS=windows go vet`, root build, and `desktop/vpnservice` — all green.

openrung-admin:
All seven items addressed in 22afb64.

**Correctness:**

1. **DNS resolution protected.** wsscore gains `ProtectedResolver` — `net.Resolver{PreferGo: true}` whose query sockets dial through `SocketControl` — wired into wsscore's own WSS network dialer, brokerapi's `NewHTTPClientWithDialControl` (inline, to avoid a cross-sibling dependency), and every protected engine dialer (`protectedNetDialer` for the reachability/ranker probes and the network-alive gate, `protectedTransport` for geo/punch). **wsscore 0.6.0 → 0.7.0**, pins moved in connectcore/root/desktop/desktop-volunteer. `TestProtectedResolverRoutesQuerySocketsThroughTheProtector` proves a resolver query socket reaches `Protect` and that the WSS dialer carries the resolver (and that a nil protector keeps the platform resolver). One caveat worth noting for Track B: `PreferGo` needs a nameserver source (`/etc/resolv.conf`), which Android lacks — the binding will need to supply platform DNS (or rely on libbox's resolver); that is a B1 wiring question, but the protection gap itself is closed and the design is self-consistent on iOS/desktop.
2. **Directory listings ride the protected client.** `identityForDirectory` now carries `HTTPClient: s.brokerHTTPClient()`; `TestDirectoryDiscoveryUsesTheProtectedBrokerClient` also pins that it's the *same pooled client* as the connect path, and nil without a protector.
3. **Refusal no longer reads as "network down".** `networkAlive` short-circuits on `isSocketProtectionFailure` and reports alive — the gate can't measure the network and waiting can't help, so the recovery ladder runs and lands the terminal local failure instead of holding CONNECTING forever. `TestNetworkAliveReportsAliveOnProtectionRefusal` covers both directions.

**Hardening:**

4+5. All three protector-derived clients (broker, geo, punch) now live in one `protectedHTTPClients` cache keyed by the protector (recover-guarded interface compare, same idiom as `sameWriter`): pooled connections are reused across calls, and a changed protector rebuilds the set and releases the old pools — an Engine kept across a VpnService recreate is never latched to a dead service. `TestProtectedClientsCachePerProtectorAndRebuildOnChange`.
6. The heartbeat fork is gone: `clienttelemetry.RunHeartbeatLoopGated(ctx, interval, gate)` carries the two host knobs (test-interval override, pre-send gate), `RunHeartbeatLoop` delegates to it, the engine passes `awaitResumed`, and the `NextHeartbeatDelay` export (added by this PR) is removed. One loop, no path future policy changes can miss.
7. `protectorDialControl` and the private sentinel are gone: wsscore exports `SocketControl` (its internal adapter now delegates to it), and classification is a single `errors.Is` on `wsscore.ErrSocketProtectionFailed`.

Re-ran everything: wsscore + brokerapi + connectcore full race suites, root with release tags (race), desktop/vpnservice + desktop-volunteer, `GOOS=windows` vet for connectcore and wsscore, and both Windows cross-builds — all green.

openrung-admin:
All eight round-4 findings addressed in 9c7a529. The two blockers were verified empirically before fixing (a standalone repro confirmed `errors.Is` = false with the sentinel visible only inside `DNSError.Err`).

**Blockers:**

1. **Refusal survives the DNSError boundary.** New `wsscore.IsSocketProtectionFailed` matches the sentinel through `errors.As(*net.DNSError)` + substring on `DNSError.Err` (the net package stringifies the resolver dial's cause, severing the wrap chain). wsscore's own `DialClient` classification and the engine's `isSocketProtectionFailure` both use it. `TestIsSocketProtectionFailedSurvivesTheDNSErrorBoundary` drives a real refused hostname dial through a real resolver, and the engine-side `TestNetworkAliveReportsAliveOnProtectionRefusal` now covers a hostname front — the exact shape that previously wedged recovery.
2. **The protected resolver requires host nameservers.** `ProtectedResolver(protector, nameservers)` returns nil without them — the platform resolver is kept (working, connection socket still protected), so re-pinning wsscore 0.7.0 cannot regress mobile's current CDN dials. With them (Android LinkProperties → new `Engine.SetDNSServers`, `ClientOptions.DNSServers`, brokerapi's `dnsServers` variadic), the pure-Go resolver queries the supplied servers — ignoring the resolv.conf-derived address — through protected sockets. The DNS-server set is part of the client-cache key, so a network change that brings new DNS rebuilds the resolvers.

**Correctness:**

3. **Epoch baseline at attempt start**, threaded through `candidateResult.netEpoch` and adopted by promote: a handover between the transport's dial and promote is now still pending after CONNECTED and rebuilds the session. The test that enshrined the absorb (`TestNetworkEpochBeforePromoteDoesNotRetireFreshSession`) is replaced by `TestEpochBetweenDialAndPromoteRetiresTheFreshSession`, which pins both directions — pre-attempt epochs stay absorbed, a mid-attempt one retires.
4. **Synchronized protector swap**: `SetSocketProtector`/`SetDNSServers` under a dedicated mutex, all engine reads via `currentProtector()`, field doc forbids post-Start writes; `TestSetSocketProtectorIsSafeMidSession` churns the swap against a live session under `-race`.
5. **Single probe on resume**: `timer.Reset` moved after the pause gate; `TestResumeAfterLongPauseRunsOneProbeNotTwo` asserts the post-resume probe spacing.

**Cleanups:** the nil-unsafe `socketControl` wrapper is deleted (inline adaptation of the nil-safe `SocketControl`, with a nil-protector assertion added to its test); `ensure(nil)` now releases and clears the cached client set instead of pinning the dead protector's pools; the callerless `FlushOnShutdown` export is gone (grep-verified: no callers in this repo or the mobile repo).

Re-ran everything: wsscore, brokerapi, connectcore (race, full), root with release tags (race), desktop/vpnservice, desktop-volunteer, `GOOS=windows` vets for connectcore and wsscore, both Windows cross-builds — all green.

openrung-admin:
All three cleanups in the follow-up commit:

1. **One resolver implementation.** `NewHTTPClientWithDialControl(timeout, control, resolver *net.Resolver)` — brokerapi's line-for-line copy of the protected-DNS logic is deleted; the engine passes `wsscore.ProtectedResolver` prebuilt (it already imports wsscore), so the censorship-critical resolver behavior exists exactly once.
2. **No forked literals.** The geo client's timeout is the newly exported `clienttelemetry.GeoLookupHTTPTimeout` (the same constant `LookupGeoAttributes` defaults with), and `punchHTTPClient(insecure, transport)` is now the single owner of the `InsecureSkipVerify` hub TLS config and the hub timeout — the engine's protected path hands it a transport instead of re-building either. Both lockstep comments are gone because there is nothing left to keep in lockstep.
3. **Refusal classification at the marking chokepoint.** `markDirectPathError`/`markWSSTransportError` detect the sentinel themselves and return the `socket_protection` local error; the five hand checks collapse to the two non-marking consumers that need their own semantics — the ticket loop's early stop (don't grind the remaining broker fronts) and `networkAlive`'s fail-alive escape. A future dial path gets the classification for free by using the markers, which everything already does.

Behavior unchanged; the existing refusal regression suite (dial, ticket, bridge-dial, hostname-through-resolver) passes against the relocated classification. Re-ran connectcore + brokerapi race suites, root `cmd/client`/`internal/tunnel` with release tags, `desktop/vpnservice`, and the windows vet — all green.

openrung-admin:
**Review follow-ups — deferred to a future separate PR** (findings 4–6 from the third review pass; findings 1–3 were fixed in abcf6e5, correctness is clean).

### 1. Socket-protection coverage is convention, not mechanism (`connectcore/sockets.go`)

Every dial site must individually remember to route through `brokerHTTPClient` / `protectedNetDialer` / `ClientOptions`, and this PR itself had to hunt down ~7 scattered sites. Nothing (no wrapper-only API, vet rule, or test sweep) forces a future engine dial onto the protected path. The failure mode is silent by design: a new broker endpoint or probe written with a bare `net.Dialer` / `http.Get` works on desktop and in every test, then silently routes into the live tunnel and blackholes connect/recovery only on Android with a VpnService up.

**Suggested fix:** an engine-owned dialer/client factory as the single sanctioned entry point, with bare `net.Dialer` / `http.Client` constructions flagged inside connectcore — make the safe path the only path.

### 2. The protector is tracked in three places (`connectcore/sockets.go:63`)

The exported `SocketProtector` field, `protectorOverride`, and a `protectorReplaced` bool that only distinguishes "setter never called" from "set to nil" — when `SetSocketProtector` could simply write one field under `protectorMu` (all reads already go through `currentProtector`). Every future reader must reconstruct the two-source-of-truth rule (field before Start, override after), and a direct write to `s.SocketProtector` after `SetSocketProtector` is silently ignored.

**Suggested fix:** collapse to one locked field; pre-Start assignment keeps working, and the flag, the override, and the trap all disappear.

### 3. `RunHeartbeatLoopGated` exports a test-only interval parameter (`connectcore/clienttelemetry/manager.go:189`)

The sole production caller passes `s.heartbeatTick`, which nothing outside connectcore's tests ever sets — yet a shared, versioned module's public API now carries a cadence override whose doc says it exists so policy "cannot fork". Future hosts can start passing their own interval, forking the randomized Android cadence the comment protects.

**Suggested fix:** inject the test cadence the way `healthTick` already works engine-side (or an unexported hook), keeping the randomized cadence the single public behavior and the exported surface one parameter smaller.

PR #171, opened 2026-08-30
ADR-003 A3: absorb the mobile policy richness and the D4 outbox debt

openrung-admin:
## What

Track A, PR A3 of ADR-003: the D4 outbox policy moves from the mobile binding into `connectcore/clienttelemetry` (the one known Go-vs-Go duplication), and the Kotlin/Swift policy trio — `WssFallbackPolicy`, `WssTicketClient`, `PunchRecoveryCircuitBreaker` — was diffed branch-by-branch against the engine, with everything the engine lacked upstreamed under tests naming the platform sources. `connectcore` 0.3.0 → **0.4.0**.

## 1. The outbox absorption

`clienttelemetry.Outbox` is the binding's `openRungTelemetryOutbox` ported whole — the NDJSON file, enqueue cap with oldest-first eviction, identity-homogeneous upload batches under the broker's 100k per-application flow budget, heartbeat piggybacking gated on the heartbeat's own identity, and all five #97 hardenings: validated element-by-element legacy-array migration, fsync durability for imports whose only other copy the caller deletes, torn-tail repair, the advisory cross-process lock, and the load-failure latch. Its regression suite moved with it (17 tests, including the 0.3.5-regression end-to-end backlog test, byte-for-byte scenarios). Two additions the binding never needed:
- **Windows lock/sync implementations** (`LockFileEx` fail-immediately; dir-sync a documented no-op) — connectcore, unlike the binding, also runs in the desktop app.
- A typed, ctx-based API: the gomobile JSON-string boundary, platform posting headers, and the cancelable single-use upload wrapper stay in the binding as its thin adapter (**the paired mobile PR reduces `telemetry_binding.go` to adapter-only and deletes the moved tests**).

The engine adopts it end to end: `Manager.UsePersistentOutbox` routes Record/Flush/Heartbeat and the geo back-patch (mobile's `ApplySessionAttributes`) through the store, and `Engine.TelemetryOutboxDirectory` opens ONE shared outbox for all sessions. `TestEngineTelemetryOutboxCarriesStrandedEventsAcrossEngines` proves the point of it all: a session whose flush never reached the broker (dead broker + 200ms Shutdown budget) is uploaded whole — `connection_ended` included — by the next engine on the same directory. Shutdown's doc now states the durable behavior it promised for A3.

## 2. Parity table

Sources inventoried at exact file:line from the mobile main tree (Kotlin `android/app/src/main/java/com/openrung/...`, Swift `ios/Shared` + `ios/PacketTunnel`).

### WssFallbackPolicy (.kt / .swift) — engine was already at parity

| Mobile branch | Engine home | Test |
|---|---|---|
| Eligibility: transport empty-or-direct, foundation, exit direct, port == 443, fronts non-empty | `supportedWSSFronts` | `TestSupportedWSSFrontsRequiresCanonicalEligibleRelay` |
| Validator exactness (normalized == signed list, else ineligible) | `supportedWSSFronts` (wsscore.NormalizeFronts equality) | same test ("reordered", "bad URL") |
| Only a direct-path failure unlocks fallback; local errors never do | `directPathError` gating in `attemptCandidate` | `TestWSSLocalSetupFailuresDoNotRequestTicketsOrDamageHealth` |
| Empty fronts → re-lock, original direct failure surfaces | `attemptCandidate` | `TestTUNModeSkipsWSSFallback` + ladder tests |
| Sequential signed-order fronts, one pass, local error aborts ladder mid-pass | front loop in `attemptCandidate` | `TestWSSFrontFailoverUsesExactAdvertisedFrontAndDoesNotAddRelayDamage` |
| Exhausted marker walks wrappers (`relayFailureAlreadyRecorded`) | `markRelayFailureRecorded`/`errors.As` | `TestWSSDirectFailureFallsBackOnSameRelayAndPreservesReality` |
| No fallback state across attempts/epochs | stateless per-ladder | `TestNetworkEpochRetiresWSSSessionAndRecoversDirectFirst` (direct-first recovery) |

### WssTicketClient (.kt / .swift) — three behaviors upstreamed ⬅ new

| Mobile branch | Engine home | Test |
|---|---|---|
| **Total 15s deadline over fronts + rounds + wait; attempts get min(5s, remaining); no attempt at remaining ≤ 0** | `wssTicketTotalDeadline` in `requestWSSSessionTicket` ⬅ new | `TestWSSTicketLadderHonorsTheSharedTotalDeadline` |
| **Wait-fits gate: a Retry-After ≥ remaining is not taken** | ⬅ new (and it leaves the once-per-ladder retry unconsumed) | `TestWSSTicketRetryWaitMustFitTheRemainingBudget` |
| **First-error-wins across BOTH rounds** | firstErr hoisted out of the round loop ⬅ new | `TestWSSTicketFirstErrorWinsAcrossRounds` |
| 2 rounds; 429/503 only; Retry-After > 0 else 10s default; 30s cap; max-of-candidates | pre-existing (`wssRetryAfter`) | `TestWSSTicketBrokerFailoverAndBoundedRetryAfter` (now with an explicit budget, since a 30s wait can never fit the 15s default — true on mobile too) |
| Per-attempt 5s budget; front dedup; cancellation aborts | pre-existing | same + `TestWSSTicketRequestsExcludeEndpointUnboundAzureFronts` |
| Ticket URL binding + strict expiry (no skew) | `attemptWSSCandidate` (ticket_binding / ticket_expired) | `TestWSSDirectFailureFallsBack…` fixture helpers |
| Local platform failure aborts the ticket ladder | protection-refusal short-circuit (A2) | `TestWSSTicketProtectionRefusalStopsAfterOneAttempt` |

### PunchRecoveryCircuitBreaker (.kt / .swift) — ported whole ⬅ new (`punchbreaker.go`)

| Mobile branch | Test |
|---|---|
| Backoff ladder: 2s→4s→8s→…→30s nominal, jump-to-cap past half, ±20% jitter, clamped (the platform tests' 100/250 → [80,160,200,200,200]) | `TestPunchBreakerBackoffLadderMatchesTheMobileTests`, `TestPunchBreakerDefaultLadderWindows` |
| Stable-path reset: uptime ≥ 5min resets streak+circuit BEFORE the loss counts; threshold−1 does not | `TestPunchBreakerStablePathResetBoundary` |
| Exempt loss: no count, no delay; stable reset still applies | `TestPunchBreakerExemptLossCountsNothingButStillResets` |
| Opens at ≥ 3 counted rapid failures, carrying that failure's backoff; per-relay; clock regression → 0 uptime | `TestPunchBreakerOpensAtThreeRapidFailuresPerRelay` |
| Engine integration: markDirectConnected at punched promote; loss decision in supervise; `punch_skipped` reason `recovery_circuit_open`; `punch_fallback` with `rapid_failure_count`/`direct_uptime_ms`/`recovery_delay_ms`; backoff before re-ladder; reset = fresh user connect (per-connection breaker) | `TestPunchFlappingOpensTheRecoveryCircuitAndFallsBackToTheHub` |

## 3. Resolved divergences

- **Kotlin-vs-Swift splits, engine's side chosen:** breaker failure exemption follows **Android** (active network probe — the engine's `networkAlive` gate — not iOS's passive NWPath snapshot); breaker clock is Go's monotonic time (suspend behavior platform-defined; Android counts sleep, iOS doesn't); `punch_fallback`'s `failure_detail` uses the engine's **telemetry-safe rendering** instead of mobile's raw `reason.take(256)` — raw error text can carry local paths, which never reach the broker.
- iOS's epoch-triggered punch exemption has no engine analog by design: an epoch never retires a punched path directly (A2's kick-then-probe), so the case reduces to the network-down exemption.
- The engine keeps its **once-per-ladder** Retry-After wait (mobile budgets per ticket call) and its endpoint-unbound front filtering — both deliberate, both pre-existing.
- Outbox `EnqueueBatch` takes typed events; JSON-array decoding stays in the binding adapter (undecodable elements can't reach the core).

## Also

- Folds in the A1 review nit: an aborted launch unwinds with the candidate's mode-appropriate grace instead of the 5s default (`subprocessRuntime.abortGrace`).

## Tests run (macOS)

connectcore `go vet` + `go test -race ./...` (all packages) + `GOOS=windows go vet ./...`; root module vet (plain + release tags) + `go test -race` with release tags; desktop/vpnservice + desktop-volunteer race suites; Windows cross-builds of cmd/client and the desktop app. Not run: real-Windows suites (CI's go-windows job; the new Windows outbox lock compiles under `GOOS=windows go vet` but executes only in CI or on a Windows box) and the mobile repo (the paired adapter PR follows after the tag).


openrung-admin:
All ten findings addressed in 17f8433 — eight fixed, two resolved as documented contracts (with reasoning below).

**Fixed:**

1. **TUN health triggers no longer count outages as punch instability** (`monitor.go`): the counting gate now probes the network itself for TUN health triggers — `(triggerWasHealth && !s.tunMode()) || networkAlive(...)` — since the TUN health loop deliberately skips its own gate. The probe runs post-teardown, over the restored routes, which is exactly Android's `physicalNetworkAlive()` placement. `TestTUNHealthLossesDuringNetworkOutageAreExemptFromThePunchBreaker` drives five health-triggered punched losses through a simulated outage: no circuit, no `punch_fallback`, and the exempt path takes the network-recovery wait.
2. **Record falls back to memory** on a store refusal, and **Flush drains both queues** — pre-attach events included — skipping (not failing on) a store another process owns. `TestManagerFallsBackToMemoryWhenTheOutboxIsUnavailable` covers the locked-out store, the pre-attach event, and delivery of both.
3. **Truncated legacy arrays salvage their decodable prefix** via streamed element decoding instead of the whole-array `Unmarshal` — which was the mobile binding's own bug, faithfully ported; the paired mobile PR inherits this fix. `TestOutboxTruncatedLegacyArraySalvagesThePrefix`.
4. **Poison head batches are discarded**: new `ErrBatchRejected` sentinel — the send function wraps it around a definitive broker refusal (the engine's send classifies 4xx except 408/429), and `FlushNextBatch` drops that batch while surfacing the error; a rejected heartbeat piggyback discards nothing, since the rejection may be the heartbeat's own (the next flush classifies the head batch alone). `TestOutboxDiscardsAPermanentlyRejectedHeadBatch`.
5. **`syncOutboxFile` opens for write** — FlushFileBuffers on Windows requires it; the O_RDONLY sync would have failed every Windows legacy import.
6. **All outbox file opens go through the sudouser hardening** the sibling identity store already uses (`OpenRegularFile` + `ChownFile`): O_NOFOLLOW-class opens, ownership handed back under sudo — an elevated TUN run can no longer root-own the telemetry state out from under later plain runs.
7. **The recovery-wait log** is transport-neutral now that punched-path exempt losses reach it.
8. Minor: the unavailable-store path in `Flush` distinguishes `ErrOutboxUnavailable` (fall through to memory) from transport errors (surface).

**Resolved as documented contracts:**

9. **`Enqueue`'s boolean** now has a precise doc: true = accepted (queued; persisted best-effort with the rewrite fallback), false = the event is nowhere and may be re-queued elsewhere. Returning the append outcome instead would break the manager fallback: an append-failed event stays in the outbox's memory and remains uploadable, so a false there would make the manager double-queue it. Callers needing durable acceptance use `EnqueueBatch`, whose contract is fsync-backed — that split is the binding's original design, now stated instead of implied.
10. **The wait-fits gate is the shipping mobile policy**, kept deliberately and now documented in place: with the 15s shared budget, a Retry-After hint beyond the remaining budget (including 30s-clamped ones the engine previously slept through) surfaces the failure and moves the ladder on. Common 429s (the ≤10s default hint) still wait and retry as before; only oversized hints change behavior — and the once-per-ladder retry stays unconsumed for a later rung whose hint fits. Mobile has run exactly this gate in production since 0.3.x; diverging here would forfeit the parity A3 exists to establish.

Re-ran: connectcore full race suite (+ `GOOS=windows` vet), root `cmd/client` with release tags, `desktop/vpnservice`, Windows client cross-build — all green.

openrung-admin:
Both follow-up findings verified and fixed in the latest commit:

1. **Byte-budgeted batch selection.** `outboxUploadBatch` now enforces a serialized-size budget (`brokerapi.MaxTelemetryBodyBytes` minus envelope slack; the heartbeat's own marshaled size is subtracted on the piggyback path) alongside the event-count and flow budgets, stopping the batch at the boundary. An event that alone exceeds the budget — or cannot marshal — is unsendable however batched, so it is discarded rather than retained as the permanent head of every future selection. `TestOutboxBatchSelectionHonorsTheBodyByteBudget` queues 200 ~4KB events plus one budget-sized whale: every request stays under the cap, the queue splits into multiple requests, all sendable events arrive, the whale never reaches the wire, and nothing wedges.

2. **Ordered fallback recovery.** `Record` now migrates the in-memory fallback into the store, oldest first, before any newer event lands there — so either the whole stream sits in the store in order, or the new event queues behind its elders in memory (and an invalid fallback event, unsendable by any queue, is dropped rather than allowed to wedge the migration). `TestManagerFallbackPreservesEventOrderAcrossRecovery` reproduces the exact scenario: record during lock-out, release the lock, record again — delivery order is first, then second.

Re-ran the full connectcore race suite, `GOOS=windows` vet, root build, and desktop/vpnservice — all green.

openrung-admin:
Verified and fixed in the latest commit: `Heartbeat` now runs the same fallback migration as `Record`, under the same lock, before touching the store — so when an outage ends and the next operation is a heartbeat rather than a record, the stranded memory event migrates into the store in order and (identity matching) rides that very heartbeat's piggyback instead of being delivered after it. An unfinished migration (store still unavailable) degrades exactly as before: the heartbeat goes alone by design, and the memory remainder drains through the post-heartbeat `Flush`, still behind nothing newer.

`TestManagerHeartbeatMigratesTheFallbackBeforeUsingTheStore` reproduces the exact sequence — strand during lock-out, end the outage, heartbeat as the next operation — and asserts the stranded event ships ahead of the heartbeat with the queue fully drained. One nuance the test surfaced: a *historical* (different-identity) store head still never rides a piggyback, so the stranded event's in-order delivery via the heartbeat depends on the identity match — which is the mobile piggyback policy working as specified, not a gap; a mismatched head drains through the follow-up Flush in stream order.

Full connectcore race suite green.

openrung-admin:
Verified and fixed in the latest commit: `Heartbeat` now consults `migrateFallbackLocked`'s answer — a completed migration takes the store path as before, while an incomplete one (the store still held by another process) falls through to the in-memory piggyback path, which ships the stranded events ahead of the heartbeat in the same request. Order preserved, cadence undelayed, exactly as suggested.

`TestManagerHeartbeatUsesTheMemoryPiggybackWhileTheStoreIsHeld` reproduces the repro's shape with the lock held for the whole test: wire order is `[stranded, session_heartbeat]` and the memory queue is committed by the piggyback. Full connectcore race suite green.

PR #174, opened 2026-09-01
ADR-003 A4: event-sequence contract vectors

openrung-admin:
ADR-003 Track A, step 4 (final): freeze the engine state machine's observable behavior as data, the way `contract/vectors` froze the token taxonomy — so the Track B bound-engine and platform-adapter suites have a machine-checkable definition of "behaves like the engine" before any of them exist.

## What this adds

**`contract/vectors/event_sequence.json`** — a new golden vector file. Each scenario scripts:

- an **initial relay directory** (identities only — WSS front ids, punch capability; addresses stay the driver's business),
- **canonical transport outcomes** (`dials` per relay as cause tokens, a failing punch establisher, a held readiness probe, a telemetry-refusing broker),
- a **timeline of host actions** (`connect`, network observations into `UpdateNetworkState`, a scripted tunnel death, `disconnect`, `Shutdown` with a flush budget, plus await sync-points),

and freezes three ordered output streams, compared exactly:

- **statuses** — engine status transitions (consecutive duplicates collapsed: the contract is transitions, not emission counts),
- **notices** — every typed notice, projected to its contract fields (wait durations excluded),
- **events** — telemetry events in broker arrival order, deduplicated by event id, `session_heartbeat` excluded, attributes projected onto the contract-bearing keys (`transport`, `from_transport`, `to_transport`, `front_id`, `trigger`, `reason`, `nat_class`, `from_relay_id`, `failure_reason`, `failure_stage`). `failure_detail`, geo, platform labels, and all measurements are host-variant metadata and deliberately outside the contract.

The full format spec lives in the file's own `format` block — the single copy, like the other vector files' doc headers.

## Scenario coverage (the ADR's acceptance list, plus one)

| # | Scenario | Freezes |
|---|---|---|
| 1 | `clean_direct_connect` | baseline session shape: preparing→connecting→connected, `connection_attempted`/`connection_succeeded`, `tunnel_stopped` + `connection_ended` teardown |
| 2 | `direct_failure_unlocks_wss_fallback` | one relay-health dent then `transport_fallback`, `wss_fallback` notice naming relay+front, success over WSS with `front_id` |
| 3 | `tunnel_death_fails_over_mid_session` | connected→connecting (never disconnected), bare `relay_attempt_failed`, one `relay_failover` (not a second success), failover notices carrying the trigger reason |
| 4 | `punch_failure_falls_back_to_hub` | `punch_attempted`→`punch_failed` (reason + NAT class), `punch_outcome` notice, silent hub fallback |
| 5 | `epoch_mid_connect_retires_fresh_wss_session` | pre-attempt epochs absorbed; a mid-attempt handover retires the fresh session orderly (`transport_session_ended` with `trigger=network_epoch`, never `transport_failed`), fresh direct-first recovery |
| 6 | `epoch_on_direct_path_probes_instead_of_retiring` | the recorded A2 divergence: an epoch kicks one health sweep, the probe stays the authority, the session survives |
| 7 | `shutdown_flushes_pending_telemetry` | the D4 debt as policy: a whole session's backlog held by a refusing broker is delivered in record order by `Shutdown`'s bounded terminal flush; no transient `disconnecting` on host shutdown |

## Generated, not hand-written

`UPDATE_SEQUENCE_VECTORS=1 go test ./connectcore -run TestEventSequenceVectors` rewrites every `expect` block from a real engine run driven through the A1 stub runtime (`TunnelRuntime`) and the A2 signal seams (`UpdateNetworkState`, `Shutdown`). Generation doesn't weaken the contract: a behavior change fails the suite, and regenerating is an edit like any other — version bump, pinned-constant bump, re-vendor. The existing `contract-vectors-version` CI job covers the new file automatically (new-on-branch files pass at their initial version).

Determinism is by construction (periodic sweeps/heartbeats suppressed, outcomes scripted, geo lookup stubbed) and was hammered: 20 plain + 5 `-race` runs, all identical.

## The `pending_suites` mechanism

The ADR asks for kotlin/swift/ts consumption declarations. Only `suites` existed, and declaring a suite with no runner is the vacuous-green problem the README warns about — so this PR adds `pending_suites`: committed future consumers, validated (known names, no duplicates, disjoint from `suites`), graduating into `suites` in the PR that wires their runner. `event_sequence.json` ships `suites: [go]`, `pending_suites: [kotlin, swift]`. There is deliberately no `ts` (classification.json's reason: the RN layer renders state, it never orchestrates a connect). Track B's punchbridge Go suite consumes the vendored copy first, against the gomobile-bound engine — a known B1 need: the bound-engine driver will want scripting seams the Go runner reaches via unexported fields; that's B1's problem to solve (likely a small exported test-support surface in a follow-up connectcore bump).

## Resolved divergences

None — this PR changes no engine behavior. Zero production-code changes beyond the `contract` package's new file constant; everything else is test code and vectors. Two frozen values worth flagging for review rather than changing:

- Scenario 3/5 freeze `failure_reason: "unknown"` on the failover records — the scripted trigger errors are deliberately unclassifiable (`cause: "unclassified"`), because the classification of a runtime's exit error is the runtime's business (a subprocess gives `process_exited`, in-process libbox gives whatever libbox reports) and belongs to `classification.json`, not the sequence contract.
- Scenario 4 freezes `nat_class: ""` on `punch_failed` — the engine records the establisher's (empty) NAT class verbatim today; freezing it keeps that visible rather than papering over it.

## Verification

- `connectcore`: full `go test ./... -race` green; the vector test additionally 20× plain + 5× `-race`, identical output every run
- root module: `go build ./... && go test ./...` green
- `desktop/`, `desktop-volunteer/`: build + full suites green
- Windows: `GOOS=windows go build ./...` (root) and `GOOS=windows go vet ./...` (root + connectcore) green
- `connectcore` VERSION 0.4.0 → 0.5.0 (all consumers use path `replace` directives; no go.mod pin moves; brokerapi/wsscore/punchcore untouched)


openrung-admin:
Review findings addressed in 0ea1819 — all eleven, each verified against the code first:

**Correctness / robustness**
- **Regen deleting unmirrored JSON fields**: real by construction (`json.Unmarshal` drops unknown fields, update mode marshals the mirror struct back). The suite now strict-decodes the file with `DisallowUnknownFields` on every run and fails naming the missing field before a regen can delete it. The guard proved itself immediately: it caught the stale `reason` fields during this very fix cycle.
- **Sink ACKing an undecodable batch**: a decode failure now returns 400, never a 204 that tells the manager to drop events the sink discarded. (The pre-existing `telemetrySink` had the same flaw; fixed in the shared sink.)
- **Geo-var swap racing the leaked lookup goroutine**: real — the lookup is abandoned, never awaited, and nothing ordered the goroutine's read of the package var against the `t.Cleanup` restore. Fixed in the engine, not the test: `attachGeoAttributes` snapshots the var on its caller's goroutine (the connect flow), giving happens-before edges on both sides of the test's swap/restore. One-line production change, behavior-neutral.
- **Unconsumed dials rows**: a `dials` entry naming a relay absent from the directory now fails the scenario up front.
- **Mis-scripted steps hanging/panicking**: `crash_tunnel` hands its error to a live run over an *unbuffered* channel with a bounded wait (a crash with no session, or a second crash for one run, is a step failure — and can no longer sit in a buffer poisoning a later run); `release_ready`/`await_ready_hold` without `hold_first_ready` are refused, as is a double release.

**Contract design**
- **Frozen display prose (json:240)**: agreed, and fixed in the reviewer's direction — `Notice.Reason` is out of the projection entirely. It is human-readable presentation (engine display strings, or a wrapped failure's `Error()` text), excluded for exactly the reason `failure_detail` is. Nothing contract-bearing is lost: the trigger distinctions ride the telemetry attributes (`transport_session_ended.trigger=network_epoch`, `punch_failed.reason=hub_declined`, the `relay_attempt_failed`-vs-`transport_session_ended` shapes). Vectors regenerated; version stays 1 (the file is still inside its introducing PR). The earlier fixed-text `scriptedError` wrapper stays as debugging hygiene, with its comment and the `causes` note rewritten to say rendered text is never part of the contract.
- **`hold_first_ready` spec mismatch (json:107)**: the spec now says what the reference runner does — every probe blocks until `release_ready`; probes reached after the release pass immediately.

**Simplification / docs**
- `sequenceTelemetrySink` folded into the shared `telemetrySink` (`hold`/`release`/`snapshot`).
- The update path derives its target from `contract.EventSequenceVectors`.
- The contract package doc now names all four contracts and the `suites`/`pending_suites` mechanism.

Verification: full connectcore `-race` suite green; the vector test 20× plain + 5× `-race` identical; root, desktop, desktop-volunteer suites green; Windows cross-build + vet (root and connectcore) green.

openrung-admin:
Round-2 findings addressed in b254612 — all ten verified first; the two biggest were dead right:

**pending_suites collision (json:7)** — confirmed by reading the mobile repo: `testdata/contract/pin.json` already owns a `pending_suites` mechanism with the opposite reading (a file declares intended consumers in `suites`; the pin records per file which declared suites the mobile repo runs — `local_suites` — vs is pending with a reason, and `contract:check` fails on a declared non-go suite that is neither). The ADR's "existing suites/pending_suites mechanism" meant *that*; I reinvented it incompatibly. Fixed by deleting my file-level field and its validation machinery entirely: `event_sequence.json` now declares `suites: [go, kotlin, swift]` — exactly `relay_decode.json`'s unwired-but-pending precedent — and the mobile pin records kotlin/swift as pending when the file is vendored. The vendoring gap is also real and now stated where it bites: the check script fetches only pinned files, so **the Track B connectcore bump must add `event_sequence.json` to pin.json's `files`** — `contract:sync` does not discover new upstream files (the file's comment and the README both say so now; the old comment's "covered by the same vendored-copy check" claim was false and is gone).

**Geo global (helpers.go:130)** — the snapshot was indeed a bandaid: its happens-before argument needed `waitIdle` before the restore, which a test failing mid-connect never reaches. Root fix: the lookup is now a per-engine seam (`Engine.lookupGeo`, resolved like the other seams), `attachGeoAttributes` takes the resolved func, and `newLadderService` stubs it for every hermetic engine — which also deleted the same racy swap-and-restore pattern from six pre-existing test sites (engine_test ×4, outboxengine_test, sockets_test). No package global is mutated anywhere anymore.

**The rest**
- `await_ready_hold` after `release_ready` now fails: a probe signals the hold gate only while unreleased, and the step also refuses to run once released.
- Update mode refuses a partial run (`-run` subtest filter → fatal instead of freezing skipped scenarios unexercised), writes via an encoder with `SetEscapeHTML(false)`, and `format` is a `json.RawMessage` — the prose-drift class is now impossible rather than detected.
- `await_status`/`await_notice` share one poll helper whose timeout prints what was actually observed (the notice side previously printed nothing).
- Regen command is fresh-clone-runnable and documented as `cd connectcore && …` in all three places (README, the file's `regenerate` field, the runner header) — `go test ./connectcore` from the root only resolves through a gitignored `go.work`, as the reviewer said.
- Determinism note names the always-alive liveness stub; the dead `networkRetryDelay` override is dropped.
- `telemetrySink` decodes the body before taking its mutex, and an undecodable batch gets 503, not 400 — agreed that 400 is precisely what the persistent outbox's poison classifier treats as permanent.

Expect blocks are regenerated (unchanged in content — only the header moved); vectors version stays 1 while the file is inside its introducing PR. Verification: full connectcore `-race` suite, vector test 20× plain + 5× `-race` identical, root/desktop/desktop-volunteer suites green, Windows cross-build + vet green.

openrung-admin:
Round-3 polish landed in 553b877 — all five, taken now precisely because zero vendored copies exist (each of the two file renames would be a cross-repo version bump after Track B):

1. **`hold_first_ready` → `hold_readiness`** — the name now says what the gate does (every probe held until `release_ready`), matching the format note.
2. **Scenario 7 de-raced by making the release engine-ordered** — the flag is now `telemetry_held_until_session_end`: the fake broker refuses every upload until a batch carries the session's `connection_ended`. Previously a fast `release_telemetry` step could race promote's still-in-flight flush and let it deliver early, so the whole-backlog-via-terminal-flush property only held on slow interleavings; now no upload before the terminal flush can ever succeed, on any interleaving. The `release_telemetry` step is gone from the vocabulary; expects unchanged.
3. **One readiness gate** — `holdReadiness(s) (held, release)` beside `newLadderService`, used by both the sequence runner and `TestEpochBetweenDialAndPromoteRetiresTheFreshSession`; the drifted inline copy is gone (the epoch test picks up the pass-through pre-check, which is behavior-neutral there).
4. **Single authoritative parse** — `contract.LoadVersionedStrict` (version pin kept, unknown fields refused, shared implementation with `LoadVersioned`); the runner's lenient-load-plus-discarded-strict-decode pair is gone.
5. **`Engine.geoLookup()` → `geoResolver()`** — no longer collides with the `geoLookup` handle type in helpers.go.

Verification: connectcore full suite + `-race` green from inside the module, vector test 20× plain + 5× `-race` identical, root/desktop/desktop-volunteer suites green, Windows cross-build + vet green. The vectors diff is exactly the two renames, the dropped step, and the prose — no expect content changed, version stays 1.

